PR #9, opened 2019-05-19
Basic Image Creation

JBorrow:
This is my attempt at implementing a basic SPH projection (scatter) using `numba`.

The new functions in `visualisation.py` allow you to make projections of a whole `SWIFTDataset`. Nothing more, nothing less -- this is not intended to be a full replacement for any rendering libraries out there. Just a little experiment.


JBorrow:
Today is a good day
```
Generating particles
Finished generating particles
Compiling
Took 1.1308925151824951 to compile.
Scattering
Took 16.765869617462158 to scatter.
Comparing with pySPHViewer
pySPHViewer took 218.46879053115845 on the same problem.
Note that pySPHViewer is running in single-threaded mode.
That makes us 13.030567188928668 x faster 😀 
```

scripts/benchmark_visualisation.py
```diff
@@ -0,0 +1,79 @@
...
+    h = 1.778_002 * h  # The kernel_gamma we use.
+
+    t = time()
+    qv = QuickView(
+        coordinates,
+        hsml=h,
+        mass=m,
+        x_size=res,
+        y_size=res,
+        r="infinity",
```
AshKelly: I can't quite remember but I think these should  be `xsize` and `ysize` and the default to 800. In that case, your code might be even faster than you think in comparison

swiftsimio/visualisation.py
```diff
@@ -0,0 +1,195 @@
...
+@jit(nopython=True)
+def pair_min(a: int32, b: int32) -> int32:
+    if a < b:
+        return a
+    else:
+        return b
+
+
+@jit(nopython=True, fastmath=True)
+def scatter(x: float32, y: float32, m: float32, h: float32, res: int) -> ndarray:
```
AshKelly: I presume the rest of swiftsimio is float32?

scripts/benchmark_visualisation.py
JBorrow: You’re right, thanks! The dangers of **kwargs, I guess

swiftsimio/visualisation.py
```diff
@@ -0,0 +1,195 @@
...
+    # Change that integer to a float, we know that our x, y are bounded
+    # by [0, 1].
+    float_res = float32(res)
+    pixel_width = 1.0 / float_res
+
+    # If the kernel width is smaller than this, we drop to just PIC method
+    drop_to_single_cell = pixel_width * 0.5
+
+    # Pre-calculate this constant for use with the above
+    inverse_cell_area = res * res
```
AshKelly: this is the number of pixels?

swiftsimio/visualisation.py
```diff
@@ -0,0 +1,195 @@
...
+            # The number of cells that this kernel spans
+            cells_spanned = int32(1.0 + kernel_width * float_res)
+
+            # Now we loop over the square of cells that the kernel lives in
+            for cell_x in range(
+                # Ensure that the lowest x value is 0, otherwise we'll segfault
+                pair_max(0, particle_cell_x - cells_spanned),
+                # Ensure that the highest x value lies within the array bounds,
+                # otherwise we'll segfault (oops).
+                min(particle_cell_x + cells_spanned, maximal_array_index),
```
AshKelly: why not use pair min since you defined it?

swiftsimio/visualisation.py
```diff
@@ -0,0 +1,195 @@
...
+            ):
+                # The distance in x to our new favourite cell -- remember that our x, y
+                # are all in a box of [0, 1].
+                distance_x = float32(cell_x) * pixel_width - x_pos
+                distance_x_2 = distance_x * distance_x
+                for cell_y in range(
+                    pair_max(0, particle_cell_y - cells_spanned),
+                    min(particle_cell_y + cells_spanned, maximal_array_index),
```
AshKelly: again, `pair_min`?

scripts/benchmark_visualisation.py
AshKelly: Yea, I've used a library that did a clever thing where they compared the **kwargs to those expected using inspection like `__code__.co_varnames` and raise an error if you passed in something unexpected

AshKelly:
I presume the images generated by pysphviewer and your in-house code look similar?

swiftsimio/visualisation.py
JBorrow: That's a good point. This should actually be `float64` as we store positions as doubles.

swiftsimio/visualisation.py
JBorrow: That's an excellent point. I have no idea how this worked before.

AshKelly:
btw, this looks good now!

JBorrow:
Thanks ash! At the moment I'm only supporting square images of the whole box, hence why there is only one DPI parameter.

Cheers again for the review!

PR #18, opened 2019-10-17
Add Py-SPHViewer Integration

JBorrow:
This provides easy (with correct units, once py-sphviewer has been updated) integration with py-sphviewer for use for rendering dark matter and stars, which is not possible with the internal renderer.


JBorrow:
@alejandrobll may be interested in this, now merged, pull request

Issue #25, opened 2019-12-02
Large spatial masks are very slow


JBorrow:
Seems that reading spatial masked data is very slow on COSMA-7 (100mb/s).

JH is unable to reproduce - could be to do with version of h5py?

JBorrow:
This appears to be partially remedied by using sorted masks.

PR #27, opened 2019-12-02
Read perf improvements

JBorrow:
Significant performance improvements for reading data using the spatial masks. Previously we were not using sorted offsets, and this meant that our data accesses were all over the place, causing really long latencies from the filesystem and an inability to use read-ahead caching. This MR:

+ Sorts the cell offsets and positions in a stable way.
+ Ensures that the routines to read data from hdf5 files use read_direct to avoid creating copies of data.
+ Cleans up some tests that were using numpy arrays instead of hdf5 file objects.

### DIFF
--- a/swiftsimio/accelerated.py
+++ b/swiftsimio/accelerated.py
@@ -89,15 +89,16 @@ def read_ranges_from_file(
     output = np.empty(output_shape, dtype=output_type)
     already_read = 0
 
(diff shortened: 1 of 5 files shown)


JBorrow:
Closes #25

JBorrow:
@sylviaploeckinger please can you provide more information? I cannot reproduce this crash.

PR #43, opened 2020-04-20
Update docs

aborissov:
I've highlighted a few portions of the docs that you should have a look at with "ALEXEI: ...", but the rest also needs checking to make sure I haven't misunderstood what the code does.

### DIFF
--- a/swiftsimio/__init__.py
+++ b/swiftsimio/__init__.py
@@ -15,8 +15,23 @@ name = "swiftsimio"
 def validate_file(filename):
     """
     Checks that the provided file is a SWIFT dataset.
+
+    Parameters
+    ----------
+    filename : str
+        name of file we want to check is a dataset
+
+    Return
+    ------
+    True
+        if `filename` is a SWIFT dataset return successfully,
+        otherwise raise exception
+
+    Raises
+    ------
+    KeyError
+        Crash if the file is not a SWIFT data file
     """
-    # ALEXEI: add param, return, examples docs
     try:
         with h5py.File(filename, "r") as handle:
             if handle["Code"].attrs["Code"] != b"SWIFT":
@@ -32,13 +47,26 @@ def mask(filename, spatial_only=True) -> SWIFTMask:
     Sets up a masking object for you to use with the correct units and
     metadata available.
 
+    Parameters
+    ----------
+    filename : str
+        SWIFT data file to read from
+    spatial_only : bool, optional
+        flag for specifing spatial masking, default to True
+
+    Returns
+    -------
+    SWIFTMask
+        empty mask object set up with the correct units and metadata
+
+    Notes
+    -----
     If you are only planning on using this as a spatial mask, ensure
-    that spatial_only remains true. If you requrie the use of the
+    that spatial_only remains true. If you require the use of the
     constrain_mask function, then you will need to use the (considerably
     more expensive, ~bytes per particle instead of ~bytes per cell
     spatial_only=False version).
     """
-    # ALEXEI: add param, return, examples docs
 
     units = SWIFTUnits(filename)
     metadata = SWIFTMetadata(filename, units)
@@ -49,8 +77,14 @@ def mask(filename, spatial_only=True) -> SWIFTMask:
 def load(filename, mask=None) -> SWIFTDataset:
     """
     Loads the SWIFT dataset at filename.
+
+    Parameters
+    ----------
+    filename : str
+        file to containing SWIFT dataset to read
+    mask : SWIFTMask, optional
+        mask to apply when reading dataset
     """
-    # ALEXEI: add param, return, examples docs
 
     return SWIFTDataset(filename, mask=mask)
 
--- a/swiftsimio/__init__.py
+++ b/swiftsimio/__init__.py
@@ -16,6 +16,7 @@ def validate_file(filename):
     """
     Checks that the provided file is a SWIFT dataset.
     """
+    # ALEXEI: add param, return, examples docs
     try:
         with h5py.File(filename, "r") as handle:
             if handle["Code"].attrs["Code"] != b"SWIFT":
@@ -37,6 +38,7 @@ def mask(filename, spatial_only=True) -> SWIFTMask:
     more expensive, ~bytes per particle instead of ~bytes per cell
     spatial_only=False version).
     """
+    # ALEXEI: add param, return, examples docs
 
     units = SWIFTUnits(filename)
     metadata = SWIFTMetadata(filename, units)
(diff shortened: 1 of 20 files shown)


swiftsimio/__init__.py
```diff
@@ -15,6 +15,22 @@
...
     Checks that the provided file is a SWIFT dataset.
+
+    Parameters
+    ----------
+    filename : str
+        name of file we want to check is a dataset
+
+    Return
+    ------
+    True
```
JBorrow, line 26: This should be `bool` as it is the return type not return value

swiftsimio/__init__.py
```diff
@@ -31,8 +47,22 @@ def mask(filename, spatial_only=True) -> SWIFTMask:
     Sets up a masking object for you to use with the correct units and
     metadata available.
 
+    Parameters
+    ----------
+    filename : str
+        SWIFT data file to read from
+    spatial_only : bool, optional
+        flag for specifing spatial masking, default to True
```
JBorrow, line 55: Flag for only spatial masking, this is much faster but will not allow you to use masking on other variables (e.g. density). Defaults to ``True``.

swiftsimio/masks.py
```diff
@@ -162,21 +208,30 @@ def constrain_mask(
...
+                [0.0, 0.1]
+            ]
+
+            These values must have units associated with them.
 
-        These values must have units associated with them.
+        Returns
+        -------
+        cell_mask : np.array[bool]
+            mask to indicate whether cells within specified spatial range
```
JBorrow, line 234: mask to indicate which cells are within the specified spatial range

swiftsimio/metadata/cosmology/cosmology_fields.py
```diff
@@ -6,6 +6,8 @@
 
 # DEPRECATED: This information is now directly read out of the snapshots.
 
+# ALEXEI: mark everything as deprecated in addition to adding some bare-bones docs
```
JBorrow: No need to docuemnt this code as it is completely unused. Can just remove these.

swiftsimio/metadata/metadata/metadata_fields.py
```diff
@@ -66,6 +67,7 @@ def generate_units_header_unpack_single_float(m, l, t, I, T) -> dict:
 
     units respectively.
     """
+    # ALEXEI: add param, return, examples docs. Improve variable names?
```
JBorrow, line 70: You're right not to document this file- it should be fairly self explanatory.

swiftsimio/metadata/unit/unit_fields.py
```diff
@@ -15,6 +15,8 @@
 # DEPRECATED: This should not be used any more by real code as we now
 # read anything directly out of the snapshots.
 
+# ALEXEI: explicitly mark functions as deprecated and add bare-bones docs
```
JBorrow: Again no need to document

swiftsimio/objects.py
```diff
@@ -20,49 +20,106 @@ class InvalidScaleFactor(Exception):
...
+    
+    This takes the expected exponent of the array that can be parsed 
+    by sympy, and the current value of the cosmological scale factor a.
 
     This should be given as the conversion from comoving to physical, i.e.
 
     r = cosmo_factor * r' with r in physical and r' comoving
 
-    Typicall this would make cosmo_factor = a for the conversion between
+    Methods
```
JBorrow, line 52: `a_factor` is not a method but in fact a property.

swiftsimio/objects.py
```diff
@@ -108,7 +165,7 @@ def __mul__(self, b):
     def __div__(self, b):
         if not self.scale_factor == b.scale_factor:
             raise InvalidScaleFactor(
-                "Attempting to subtract two cosmo_factors with different scale factors "
+                "Attempting to divide two cosmo_factors with different scale factors "
```
JBorrow, line 168: Again, nice catch :)

swiftsimio/reader.py
```diff
@@ -85,15 +102,60 @@ def get_unit_dictionary(self):
...
+        Loads the metadata 
+    get_named_column_metadata(self):
+        Loads custom metadata from named columns
+    postprocess_header(self):
+        Postprocesses local variables in header 
+    present_particle_types(self):
+        Get particle types present in snapshot
+    present_particle_names(self):
+        Get particle names present in snapshot
+    code_info(self):
```
JBorrow, line 130: These info values are not in fact methods - they are again properties and so should be documented under `Attributes`.

swiftsimio/reader.py
```diff
@@ -867,6 +998,31 @@ class __SWIFTNamedColumnDataset(object):
...
+
+        Examples
+        --------
+        For a gas particle we might be interested in the mass fractions for a number
+        of elements (e.g. hydrogen, helium, carbon, etc). In a SWIFT snapshot these
+        would be found in `field_path` = /PartType0/ElementMassFractions. The 
+        `named_columns` would be the list of elements (["hydrogen", ...]) and the 
+        variable we're interested in is the mass fraction `name` = element_mass_fraction.
+        Thus, 
+            data.gas = __SWIFTNamedColumnDataset(/PartType0/ElementMassFractions, ["hydrogen", "helium"], element_mass_fraction)
```
JBorrow, line 1021: Please format as

data.gas = __SWIFTNamedColumnDataset(
    "/PartType0/ElementMassFractions",
    ["hydrogen", "helium"],
    "element_mass_fraction"
)

swiftsimio/visualisation/projection_backends/reference.py
```diff
@@ -38,6 +38,7 @@ def scatter(x: float64, y: float64, m: float32, h: float32, res: int) -> ndarray
     for a 25-50% performance improvement. In our testing, using numpy
     floats and integers is also an improvement over using the numba ones.
     """
+    # ALEXEI: add param, return, examples docs
```
JBorrow, line 41: Please just ask users to refer to the documentation in `fast` for parameters or we can do:
```
from swiftsimio.visualisation.projection_backends.fast import scatter as scatter_fast
from swiftsimio.visualisation.projection_backends.fast import scatter_parallel as scatter_fast_parallel
```

and after the definition of `scatter` and `scatter_parallel`:

```
scatter.__doc__ = scatter_fast.__doc__
scatter_parallel.__doc__ = scatter_fast_parallel.__doc__
```

which do you think is best?

swiftsimio/visualisation/sphviewer.py
```diff
@@ -162,18 +189,30 @@ def get_camera(
...
+        ----------
+        x, y, z : float, optional
+            Cartesian co-ordinates of the object being viewed
+        r : float, optional
+            Cartesian distance to the object
+        t : float, optional
+            ?????????? TODO
+        p : float, optional
+            ?????????? TODO
+        zoom : float, optional
```
JBorrow, line 204: Unfortunately I don't know what these parameters do now either - so can we say 'see py-sphviewer documentation`?

JBorrow:
Fantastic, thank you Alexei!

PR #45, opened 2020-04-29
Write subset

aborissov:
For a given mask this will write a subset of the datasets and copy over the metadata into a new SWIFT snapshot. 

to-do list:
- docs
- mask construction from set of particle or cell ids?
- incorporate test into automated test suite

### DIFF
--- a/docs/source/masking/index.rst
+++ b/docs/source/masking/index.rst
@@ -110,6 +110,7 @@ Writing subset of snapshot
 In some cases it may be useful to write a subset of an existing snapshot to its
 own hdf5 file. This could be used, for example, to extract a galaxy halo that 
 is of interest from a snapshot so that the file is easier to work with and transport.
+
 To do this the ``write_subset`` function is provided. It can be used, for example,
 as follows
 
@@ -127,7 +128,13 @@ as follows
     sw.subset_writer.write_subset("test_subset.hdf5", mask)
 
 This will write a snapshot which contains the particles from the specified snapshot 
-with the *x*-coordinate restricted to the range [100, 1000] kpc. 
+whose *x*-coordinate is within the range [100, 1000] kpc. This function uses the 
+cell mask which encompases the specified spatial domain to successively read portions 
+of datasets from the input file and writes them to a new snapshot. 
+
+Due to the coarse grained nature of the cell mask, particles from outside this range 
+may also be included if they are within the same top level cells as particles that 
+fall within the given range.
 
 Please note that it is important to run ``constrain_spatial`` as this generates
 and stores the cell mask needed to write the snapshot subset. 
--- a/docs/source/masking/index.rst
+++ b/docs/source/masking/index.rst
@@ -104,3 +104,30 @@ belong to the masked region (in both density and spatial) are read. I.e. if I
 ask for the temperature of particles, it will recieve an array containing
 temperatures of particles that lie in the region [0.2, 0.7] and have a
 density between 0.4 and 0.8 g/cm^3.
+
+Writing subset of snapshot
+--------------------------
+In some cases it may be useful to write a subset of an existing snapshot to its
+own hdf5 file. This could be used, for example, to extract a galaxy halo that 
+is of interest from a snapshot so that the file is easier to work with and transport.
+To do this the ``write_subset`` function is provided. It can be used, for example,
+as follows
+
+.. code-block:: python
+
+    import swiftsimio as sw                                                 
+    import unyt                                                             
+    
+    mask = sw.mask("eagle_snapshot.hdf5")                                       
+    mask.constrain_spatial([
+        [unyt.unyt_quantity(100, unyt.kpc), unyt.unyt_quantity(1000, unyt.kpc)], 
+        None, 
+        None])                                   
+    
+    sw.subset_writer.write_subset("test_subset.hdf5", mask)
+
+This will write a snapshot which contains the particles from the specified snapshot 
+with the *x*-coordinate restricted to the range [100, 1000] kpc. 
+
+Please note that it is important to run ``constrain_spatial`` as this generates
+and stores the cell mask needed to write the snapshot subset. 
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -11,6 +11,7 @@ import numpy as np
 from swiftsimio import metadata, SWIFTMetadata, SWIFTUnits
 
 from swiftsimio.accelerated import ranges_from_array
+from typing import Dict
 
 
 class SWIFTMask(object):
@@ -406,7 +407,9 @@ class SWIFTMask(object):
 
         return
 
-    def get_masked_counts_offsets(self, restrict: np.ndarray) -> (dict, dict):
+    def get_masked_counts_offsets(
+        self, restrict: np.ndarray
+    ) -> (Dict[str, np.array], Dict[str, np.array]):
         """
         Returns the particle counts and offsets in cells selected by the mask
 
@@ -426,23 +429,26 @@ class SWIFTMask(object):
 
         Returns
         -------
-        dict, dict
-            dictionaries containing the particle offets and counts for each particle type
+        Dict[str, np.array], Dict[str, np.array]
+            dictionaries containing the particle offets and counts for each particle type. For example, the particle counts dictionary would be of the form 
+                {"gas": [g_0, g_1, ...],
+                 "dark matter": [bh_0, bh_1, ...], ...}
+            where the keys would be each of the particle types and values are arrays of the number of corresponding particles in each cell (in this case there would be g_0 gas particles in the first cell, g_1 in the second, etc.). The structure of the dictionaries is the same for the offsets, with the arrays now storing the offset of the first particle in the cell. 
 
         """
         if self.spatial_only:
             masked_counts = {}
-            masked_offsets = {}
+            current_offsets = {}
             cell_mask = self._generate_cell_mask(restrict)
             for part_type, counts in self.counts.items():
-                masked_counts[part_type] = counts*cell_mask
+                masked_counts[part_type] = counts * cell_mask
 
-                masked_offsets[part_type] = [0 for i in range(len(counts))]
+                current_offsets[part_type] = [0] * counts.size
                 running_sum = 0
                 for i in range(len(counts)):
-                    masked_offsets[part_type][i] = running_sum
+                    current_offsets[part_type][i] = running_sum
                     running_sum += masked_counts[part_type][i]
 
-            return masked_counts, masked_offsets
+            return masked_counts, current_offsets
         else:
-            raise("Only applies on spatial only masks")
+            raise ("Only applies on spatial only masks")
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -405,3 +405,21 @@ class SWIFTMask(object):
                 setattr(self, ptype, ranges_from_array(getattr(self, ptype)))
 
         return
+
+    def refine_metadata_mask(self, restrict):
+        if self.spatial_only:
+            refined_counts = {}
+            refined_offsets = {}
+            cell_mask = self._generate_cell_mask(restrict)
+            for part_type, counts in self.counts.items():
+                refined_counts[part_type] = counts*cell_mask
+
+                refined_offsets[part_type] = [0 for i in range(len(counts))]
+                running_sum = 0
+                for i in range(len(counts)):
+                    refined_offsets[part_type][i] = running_sum
+                    running_sum += refined_counts[part_type][i]
+
+            return refined_counts, refined_offsets
+        else:
+            raise("Only applies on spatial only masks")
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -406,20 +406,43 @@ class SWIFTMask(object):
 
         return
 
-    def refine_metadata_mask(self, restrict):
+    def get_masked_counts_offsets(self, restrict: np.ndarray) -> (dict, dict):
+        """
+        Returns the particle counts and offsets in cells selected by the mask
+
+        Parameters
+        ----------
+        restrict : np.ndarray
+            Restrict is a 3 length list that contains length two arrays giving 
+            the lower and upper bounds for that axis, e.g.
+
+            restrict = [
+                [0.5, 0.7],
+                [0.1, 0.9],
+                [0.0, 0.1]
+            ]
+
+            These values must have units associated with them.
+
+        Returns
+        -------
+        dict, dict
+            dictionaries containing the particle offets and counts for each particle type
+
+        """
         if self.spatial_only:
-            refined_counts = {}
-            refined_offsets = {}
+            masked_counts = {}
+            masked_offsets = {}
             cell_mask = self._generate_cell_mask(restrict)
             for part_type, counts in self.counts.items():
-                refined_counts[part_type] = counts*cell_mask
+                masked_counts[part_type] = counts*cell_mask
 
-                refined_offsets[part_type] = [0 for i in range(len(counts))]
+                masked_offsets[part_type] = [0 for i in range(len(counts))]
                 running_sum = 0
                 for i in range(len(counts)):
-                    refined_offsets[part_type][i] = running_sum
-                    running_sum += refined_counts[part_type][i]
+                    masked_offsets[part_type][i] = running_sum
+                    running_sum += masked_counts[part_type][i]
 
-            return refined_counts, refined_offsets
+            return masked_counts, masked_offsets
         else:
             raise("Only applies on spatial only masks")
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -20,20 +20,32 @@ class SWIFTMask(object):
 
     Methods
     -------
-    _generate_empty_masks(self)
-        Create empty masks for all particles
-    _unpack_cell_metadata(self)
-        load cell metadata into local class variables
-    constrain_mask( self, ptype: str, quantity: str, lower: unyt.array.unyt_quantity, upper: unyt.array.unyt_quantity,)
+
+
+    constrain_mask(self, ptype: str, quantity: str,
+        lower: unyt.array.unyt_quantity, upper: unyt.array.unyt_quantity)
         constrains a particle mask based on the value of a the particle quantity
-    _generate_cell_mask(self, restrict)
-        generates spatially restricted mask for cell
-    _update_spatial_mask(self, restrict, ptype: str, cell_mask: np.array)
-        updates the particle mask using the cell mask. 
+
     constrain_spatial(self, restrict)
         generates spatially constrained cell mask
+
     convert_masks_to_ranges(self)
         converts the masks to range masks so that they take up less space.
+
+    Private Methods
+    ---------------
+
+    _generate_empty_masks(self)
+        Create empty masks for all particles
+
+    _unpack_cell_metadata(self)
+        load cell metadata into local class variables
+
+    _generate_cell_mask(self, restrict)
+        generates spatially restricted mask for cell
+
+    _update_spatial_mask(self, restrict, ptype: str, cell_mask: np.array)
+        updates the particle mask using the cell mask. 
     """
 
     def __init__(self, metadata: SWIFTMetadata, spatial_only=True):
@@ -93,10 +105,16 @@ class SWIFTMask(object):
 
         with h5py.File(self.metadata.filename, "r") as handle:
             cell_handle = handle["Cells"]
-            offset_handle = cell_handle["Offsets"]
             count_handle = cell_handle["Counts"]
             metadata_handle = cell_handle["Meta-data"]
             centers_handle = cell_handle["Centres"]
+
+            try:
+                offset_handle = cell_handle["OffsetsInFile"]
+            except KeyError:
+                # Previous version of SWIFT did not have distributed
+                # file i/o implemented
+                offset_handle = cell_handle["Offsets"]
 
             # Only want to compute this once (even if it is fast, we do not
             # have a reliable stable sort in the case where cells do not
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -430,10 +430,15 @@ class SWIFTMask(object):
         Returns
         -------
         Dict[str, np.array], Dict[str, np.array]
-            dictionaries containing the particle offets and counts for each particle type. For example, the particle counts dictionary would be of the form 
+            dictionaries containing the particle offets and counts for each particle 
+            type. For example, the particle counts dictionary would be of the form 
                 {"gas": [g_0, g_1, ...],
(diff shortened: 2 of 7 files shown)


swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+        the name of the dataset we're interested in. This is the name from the
+        hdf5 file (i.e. "PartType0", rather than "gas")
+
+    Returns
+    -------
+    int
+        size of the subset of the dataset selected by the mask
+    """
+    legend = {"Cells": 0, 
+              "PartType0": mask.gas_size if hasattr(mask, "gas_size") else 0,
```
JBorrow, line 32: You should be able to use `swiftsimio.metadata.particle.particle_name_underscores` here.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+        size of the subset of the dataset selected by the mask
+    """
+    legend = {"Cells": 0, 
+              "PartType0": mask.gas_size if hasattr(mask, "gas_size") else 0, 
+              "PartType1": mask.dark_matter_size if hasattr(mask, "dark_matter_size") else 0, 
+              "PartType2": 0, 
+              "PartType3": 0, 
+              "PartType4": mask.stars_size if hasattr(mask, "stars_size") else 0, 
+              "PartType5": mask.black_holes_size if hasattr(mask, "black_holes_size") else 0}
+    for key in legend.keys():
```
JBorrow, line 38: Should be able to remove this for loop in that case.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Returns
+    -------
+    np.ndarray
+        mask for the appropriate dataset
+
+    """
+    legend = {"Cells": 0, 
+              "PartType0": mask.gas if hasattr(mask, "gas") else 0, 
+              "PartType1": mask.dark_matter if hasattr(mask, "dark_matter") else 0,
```
JBorrow, line 62: Don't use 0 for something that doesn't exist - much better to use `None`.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+    legend = {"Cells": 0, 
+              "PartType0": mask.gas if hasattr(mask, "gas") else 0, 
+              "PartType1": mask.dark_matter if hasattr(mask, "dark_matter") else 0, 
+              "PartType2": 0, 
+              "PartType3": 0, 
+              "PartType4": mask.stars if hasattr(mask, "stars") else 0, 
+              "PartType5": mask.black_holes if hasattr(mask, "black_holes") else 0}
+    for key in legend.keys():
+        if key in dataset_name:
+            return legend[key]
```
JBorrow, line 69: Might be better to go for a try/except here:

```python
particle_types_reversed = {v : k for v, k in swiftsimio.metadata.particle.particle_name_underscores}
try:
    particle_number = int(dataset_name[-1])
except ValueError:
    # Not a particle numbered dataset - tried to convert a letter to an int.
    return None

return getattr(
    mask,
    particle_types_reversed[particle_number],
    None # Default value
)

File: swiftsimio/subset_writer.py [hunk 5]
```diff
@@ -0,0 +1,172 @@
...
+    outfile : h5py.File
+        hdf5 file handle for output snapshot
+    mask : SWIFTMask
+        the mask used to define subset that is written to new snapshot
+    dataset_names : list of str
+        names of datasets found in the snapshot
+    """
+    if mask is not None:
+        for name in dataset_names:
+            if "Cells" not in name:
```
JBorrow, line 87: This is a bit of a clunky way of handling this edge case - can we do:

```python
if "Cells" in name:
    continue
```
This way we can have the other logic one level lower

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+                output_type = first_value.dtype
+                output_size = first_value.size
+                mask_size = compute_mask_size(mask, name)
+                if output_size != 1:
+                    output_shape = (mask_size, output_size)
+                else:
+                    output_shape = mask_size
+    
+                dataset_mask = get_mask_label(mask, name)
+                subset = read_ranges_from_file(infile[name], dataset_mask, output_shape = output_shape, output_type = output_type)
```
JBorrow, line 99: Please format file with `black`

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    if mask is not None:
+        for name in dataset_names:
+            if "Cells" not in name:
+                # get output dtype and size 
+                first_value = infile[name][0]
+                output_type = first_value.dtype
+                output_size = first_value.size
+                mask_size = compute_mask_size(mask, name)
+                if output_size != 1:
```
JBorrow, line 93: You can probably use `infile[name].ndim` here instead?

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Parameters
+    ----------
+    infile : h5py.File
+        hdf5 file handle for input snapshot
+    outfile : h5py.File
+        hdf5 file handle for output snapshot
+    """
+    for field in infile.keys():
+        if not any([group_str in field for group_str in ["PartType"]]):
```
JBorrow, line 116: I'm confused by this for loop - can you not just do:

```
for field in infile.keys():
    if not "PartType" in field:
        infile.copy(field, outfile)

File: swiftsimio/subset_writer.py [hunk 9]
```diff
@@ -0,0 +1,172 @@
...
+    Parameters
+    ----------
+    input_file : h5py.File
+        hdf5 file handle for snapshot
+    dataset_names : list of str
+        names of datasets found in the snapshot
+    path : str, optional
+        the path to the current location in the snapshot
+    """
+    if path != None:
```
JBorrow, line 132: Never check for equality with `None` like this, always use `is not None` - see https://stackoverflow.com/questions/2209755/python-operation-vs-is-not

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+        keys = input_file[path].keys()
+    else:
+        keys = input_file.keys()
+        path = ""
+
+    for key in keys:
+        subpath = path + "/" + key
```
JBorrow, line 139: Please don't use addition on strings - always use formatting like `subpath = f"{path}/{key}"`

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+        if isinstance(input_file[subpath], h5py.Dataset):
+            dataset_names.append(subpath)
+        elif input_file[subpath].keys() != None:
+            find_datasets(input_file, dataset_names, subpath)
```
JBorrow, line 143: Nice use of recursion!

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,172 @@
...
+    infile = h5py.File(input_file, "r")
+    outfile = h5py.File(output_file, "w")
+
+    # Find the datasets
+    dataset_names = []
+    find_datasets(infile, dataset_names)
+    
+    # Write metadata and data subset
+    write_metadata(infile, outfile)
+    write_datasubset(infile, outfile, mask, dataset_names)
```
JBorrow, line 168: Can we be cheeky here and do `write_datsubset(infile,outfile,mask,find_datasets(infile))` with dataset_names a default parameter of `[]` in the function declaration above? I suppose this would require the addition of a return at the end of the for loop?

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+    for i in range(len(A)):
+        try:
+            if A[i] != B[i]:
+                return False
+        except:
+            if not all(A[i][:] == B[i][:]):
+                return False
+
+    # All good
+    return True
```
JBorrow, line 37: I'm confused here why `(A == B).all()` is not used - am I missing something?

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    # Initialise a list to store fields that differ
+    bad_compares = []
+
+    # Compare metadata
+    if A.metadata.time != B.metadata.time:
+        bad_compares.append("metadata")
+
+    # Compare datasets
+    possible_part_types = ['gas', 'dark_matter', 'stars', 'black_holes']
```
JBorrow, line 60: Please use the global metadata here that includes the boundary particles too.

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+    part_types = [attr for attr in possible_part_types if hasattr(A, attr)]
+    for j in range(len(part_types)):
```
JBorrow, line 62: Can we just do for `part_type in part_types` here?

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+        A_type = getattr(A, part_types[j])
+        B_type = getattr(B, part_types[j])
+        attrs = [attr for attr in dir(A_type) if not attr.startswith('_')]
+        for i in range(len(attrs)):
```
JBorrow, line 66: Could use filter here and be more concise - 

```python
for attr in filter(lambda x: not x.startswith("_"), dir(A_type)):
```

tests/subset_write_test.py
JBorrow: (same comment as below about `filter` also applies here)

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+            if not callable(getattr(A_type, attrs[i])):
+                print("comparing ", attrs[i])
```
JBorrow, line 68: `print` statements should be removed before merging

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Compare datasets
+    possible_part_types = ['gas', 'dark_matter', 'stars', 'black_holes']
+    part_types = [attr for attr in possible_part_types if hasattr(A, attr)]
+    for j in range(len(part_types)):
+        A_type = getattr(A, part_types[j])
+        B_type = getattr(B, part_types[j])
+        attrs = [attr for attr in dir(A_type) if not attr.startswith('_')]
+        for i in range(len(attrs)):
+            if not callable(getattr(A_type, attrs[i])):
```
JBorrow, line 67: Don't call getattr() twice, best to do:

```
parameter = getattr(A_type, attr)
```

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+                print("comparing ", attrs[i])
+                if not compare_arrays(getattr(A_type, attrs[i]), getattr(B_type, attrs[i])):
+                    bad_compares.append(part_types[j] + " " + attrs[i])
+
+    if bad_compares != []:
+        print("compare failed on ", bad_compares)
+    else:
+        print("compare completed successfully")
+
+    assert(bad_compares == [])
```
JBorrow, line 77: You've done the same thing twice here - why not do

```
assert bad_compares == [], f"Failure to compare on {bad_compares}"
```

tests/subset_write_test.py
```diff
@@ -0,0 +1,110 @@
...
+    
+    # Write the subset
+    write_subset(filename, outfile, mask)
+    
+    # Compare written subset of snapshot against corresponding region in full snapshot
+    snapshot = sw.load(filename, mask)
+    sub_snapshot = sw.load(outfile)
+    
+    # First check the metadata
+    compare(snapshot, sub_snapshot)
```
JBorrow, line 107: Should delete the file after you're finished with it

tests/subset_write_test.py
aborissov: I tried this, didn't work. I think this is because of the datasets which contain an array of arrays.

swiftsimio/subset_writer.py
aborissov: sorry, this was a relic of a previous version where I was also checking against "Cells". changed to the sensible loop you suggest

swiftsimio/subset_writer.py
aborissov: Tried this, didn't work. I'm looking for the _size variable (like mask.gas_size) here. I've reworked this function and the one slightly below into a single one which will select the correct attribute.

aborissov:
I've gone through all the comments now (I'm pretty sure,  I was doing them in a slightly non sequential way so might have missed one). Let me know if there's anything else that jumps out at you. Otherwise I'm happy to merge.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Returns
+    -------
+    np.ndarray
+        mask for the appropriate dataset
+
+    """
+    if "Cells" in dataset_name:
+        return None
+    elif "PartType" in dataset_name:
```
JBorrow, line 41: Why the double condition here?

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,159 @@
...
+        part_type = [int(x) for x in filter(str.isdigit, dataset_name)][0]
+        mask_name = metadata.particle_types.particle_name_underscores[part_type]
+        return (
+            getattr(mask, f"{mask_name}{suffix}") if hasattr(mask, mask_name) else None
```
JBorrow, line 45: You can provide None as the third attribute to `getattr` and it will return that if it does not find the attribute. You are also not checking for a consistent thing here.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,159 @@
...
+    Returns
+    -------
+    np.ndarray
+        mask for the appropriate dataset
+
+    """
+    if "Cells" in dataset_name:
+        return None
+    elif "PartType" in dataset_name:
+        part_type = [int(x) for x in filter(str.isdigit, dataset_name)][0]
```
JBorrow, line 42: Doing `[0]` here is probably not the best way of resolving this issue. I think it's best that we just convert the -1th element in the string, so `part_type = int(dataset_name[-1])` - this is much simpler and returns the same thing.

tests/subset_write_test.py
JBorrow: That should work - the maintainers of numpy are a lot smarter than us :).

Can you provide an example that does not work?

tests/subset_write_test.py
```diff
@@ -0,0 +1,101 @@
...
+    Compares two arrays or arrays of arrays
+
+    Parameters
+    ----------
+    A, B : np.array
+        Arrays to compare
+
+    Returns
+    -------
+    bool
```
JBorrow, line 22: Please remove this function and just use the contentss in the appropriate place

swiftsimio/subset_writer.py
aborissov: Doing `part_type = int(dataset_name[-1])` doesn't work because some of the datasets have the form PartType0/Coordinates

swiftsimio/subset_writer.py
aborissov: Don't know what I was thinking. Changed this to a single if.. else statement

swiftsimio/subset_writer.py
aborissov: Done.

JBorrow:
I just tried this on a SWIFT snapshot and it failed:

```python
import swiftsimio as sw                                                 
import unyt                                                             

mask = sw.mask("eagle_0011.hdf5")                                       
mask.constrain_spatial([[unyt.unyt_quantity(100, unyt.kpc), unyt.unyt_quantity(1000, unyt.kpc)], None, None])                                   

sw.subset_writer.write_subset("eagle_0011.hdf5", "test_subset.hdf5", mask) 
```

on the snapshot in

`/snap7/scratch/dp004/dc-borr1/density-drift-move-real/SPHENIX_UPDATES`.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+from swiftsimio.masks import SWIFTMask
+from swiftsimio.accelerated import read_ranges_from_file
+import swiftsimio.metadata as metadata
+
+import unyt
+import h5py
+import numpy as np
+
+
+def get_dataset_mask(mask: SWIFTMask, dataset_name: str, suffix=""):
```
JBorrow, line 16: Please add type annotation for suffix. Also not comfortable using a string here as the default, as that's mutable.

So:

```
suffix: Optional[str] = None
```
You can get `Optional` from the `typing` module.

Then in the code, we can do `suffix = "" if suffix is None else suffix`

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+    """
+    if "PartType" in dataset_name:
+        part_type = [int(x) for x in filter(str.isdigit, dataset_name)][0]
+        mask_name = metadata.particle_types.particle_name_underscores[part_type]
+        return getattr(mask, f"{mask_name}{suffix}", None)
+    else:
+        return None
+
+
+def write_datasubset(infile, outfile, mask: SWIFTMask, dataset_names, links_list):
```
JBorrow, line 47: Please add correct type annotations here, including for infile and outfile.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+        hdf5 file handle for output snapshot
+    mask : SWIFTMask
+        the mask used to define subset that is written to new snapshot
+    dataset_names : list of str
+        names of datasets found in the snapshot
+    links_list : list of str
+        names of links found in the snapshot
+    """
+    skip_list = links_list.copy()
+    skip_list.extend(["Cells", "SubgridScheme"])
```
JBorrow, line 64: Why not just do skip_list = links_list + ["Cells", "SubgridScheme"]

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+                output_type=output_type,
+            )
+
+            # Write the subset
+            outfile.create_dataset(name, data=subset)
+            for attr_name, attr_value in infile[name].attrs.items():
+                outfile[name].attrs.create(attr_name, attr_value)
+
+
+def write_metadata(infile, outfile, links_list):
```
JBorrow, line 95: Type annotations here please

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+        names of links found in the snapshot
+    """
+    skip_list = links_list.copy()
+    skip_list.append("PartType")
+    for field in infile.keys():
+        if not any([substr for substr in skip_list if substr in field]):
+            infile.copy(field, outfile)
+
+
+def find_datasets(input_file: h5py.File, dataset_names=[], path=None):
```
JBorrow, line 115: Type annotations including return type with `->`

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+        subpath = f"{path}/{key}"
+        if isinstance(input_file[subpath], h5py.Dataset):
+            dataset_names.append(subpath)
+        elif input_file[subpath].keys() is not None:
+            find_datasets(input_file, dataset_names, subpath)
+
+    return dataset_names
+
+
+def find_links(input_file: h5py.File, link_names=[], link_paths=[], path=None):
```
JBorrow, line 144: Note about type annotations again.

swiftsimio/subset_writer.py
```diff
@@ -0,0 +1,224 @@
...
+    links_list : list of str
+        list of names of soft links
+    paths_list : list of str
+        list of paths specifying how to link each soft link
+    """
+    for i in range(len(links_list)):
+        file[links_list[i]] = h5py.SoftLink(paths_list[i])
+
+
+def write_subset(input_file: str, output_file: str, mask: SWIFTMask):
```
JBorrow, line 199: I don't think we need `input_file` here as we can use `mask.filename`

swiftsimio/masks.py
```diff
@@ -405,3 +405,44 @@ def convert_masks_to_ranges(self):
...
+
+        """
+        if self.spatial_only:
+            masked_counts = {}
+            masked_offsets = {}
+            cell_mask = self._generate_cell_mask(restrict)
+            for part_type, counts in self.counts.items():
+                masked_counts[part_type] = counts*cell_mask
+
+                masked_offsets[part_type] = [0 for i in range(len(counts))]
```
JBorrow, line 440: `[0] * counts.size`

swiftsimio/masks.py
```diff
@@ -405,3 +405,44 @@ def convert_masks_to_ranges(self):
...
+                running_sum = 0
```
JBorrow, line 441: `current_offset` is probably a better variable name?

JBorrow:
Just tested the functionality - looks really good to me!

swiftsimio/masks.py
```diff
@@ -405,3 +405,44 @@ def convert_masks_to_ranges(self):
...
+                [0.1, 0.9],
+                [0.0, 0.1]
+            ]
+
+            These values must have units associated with them.
+
+        Returns
+        -------
+        dict, dict
+            dictionaries containing the particle offets and counts for each particle type
```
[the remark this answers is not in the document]
aborissov: Done in 909b787

swiftsimio/masks.py
```diff
@@ -405,3 +405,44 @@ def convert_masks_to_ranges(self):
...
+
+        """
+        if self.spatial_only:
+            masked_counts = {}
+            masked_offsets = {}
+            cell_mask = self._generate_cell_mask(restrict)
+            for part_type, counts in self.counts.items():
+                masked_counts[part_type] = counts*cell_mask
```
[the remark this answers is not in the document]
aborissov: Formatted the file, but it looks like black didn't do much to my docstrings, don't know if this is a "feature". Let me know if it looks ok the way I manually did it.

swiftsimio/masks.py
aborissov: Done

swiftsimio/masks.py
aborissov: Done

docs/source/masking/index.rst
```diff
@@ -104,3 +104,30 @@ belong to the masked region (in both density and spatial) are read. I.e. if I
...
+    mask = sw.mask("eagle_snapshot.hdf5")                                       
+    mask.constrain_spatial([
+        [unyt.unyt_quantity(100, unyt.kpc), unyt.unyt_quantity(1000, unyt.kpc)], 
+        None, 
+        None])                                   
+    
+    sw.subset_writer.write_subset("test_subset.hdf5", mask)
+
+This will write a snapshot which contains the particles from the specified snapshot 
+with the *x*-coordinate restricted to the range [100, 1000] kpc.
```
JBorrow, line 130: That's not true - it should say with all particles in the range [100, 1000] kpc included - we may include some extra particles and it's important to note that (as we use the top-level cell masking it's coarse grained to the top-level cell size).

JBorrow:
Fantastic, thank you @aborissov! That's all finished now.

PR #46, opened 2020-05-14
read chunks instead of ranges

aborissov:


swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+################################ ALEXEI: playing around with better read_ranges_from_file implementation #################################
+
+def get_chunk_ranges(ranges, chunk_size, array_length):
+    chunk_ranges = []
+    for bound in ranges:
+        lower = int(np.floor(bound[0]/chunk_size))*chunk_size
+        upper = min(int(np.ceil(bound[1]/chunk_size))*chunk_size, array_length)
+
+        if len(chunk_ranges) > 0:
+            assert(lower >= chunk_ranges[-1][0])
```
JBorrow, line 194: Remove asserts wherever possible - they slow down the code significantly.

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
 
     return read_ranges_from_file(handle, ranges, output_size, output_type)
+
+
+################################ ALEXEI: playing around with better read_ranges_from_file implementation #################################
+
+def get_chunk_ranges(ranges, chunk_size, array_length):
+    chunk_ranges = []
+    for bound in ranges:
+        lower = int(np.floor(bound[0]/chunk_size))*chunk_size
```
JBorrow, line 190: Please make sure that you format with black every time before you make a commit.

You should use `math` instead of `numpy` for single values - these will get compiled better by numba and should be faster than making calls out to numpy.

Aren't bound[0] and chunk_size both ints? In that case we can use integer division:

`lower = (bound[0] // chunk_size) * chunk_size`

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        upper = min(int(np.ceil(bound[1]/chunk_size))*chunk_size, array_length)
```
JBorrow, line 191: Similarly to above we can probably get away with using integer divison here. Integer division always rounds down, so you can do:

upper = min(- ((- bound[1]) / chunk_size ) * chunk_size, array_length)

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+
+        if len(chunk_ranges) > 0:
+            assert(lower >= chunk_ranges[-1][0])
+            assert(upper >= chunk_ranges[-1][1])
+            if lower > chunk_ranges[-1][1]:
+                chunk_ranges.append([lower, upper])
+            elif lower <= chunk_ranges[-1][1]:
+                chunk_ranges[-1][1] = upper
+            else:
+                raise RuntimeError("computing chunk ranges has gone horribly wrong")
```
JBorrow, line 201: Remove this before production - you should then be able to compile this function with `numba`.

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    chunk_ranges = []
+    for bound in ranges:
+        lower = int(np.floor(bound[0]/chunk_size))*chunk_size
+        upper = min(int(np.ceil(bound[1]/chunk_size))*chunk_size, array_length)
+
+        if len(chunk_ranges) > 0:
+            assert(lower >= chunk_ranges[-1][0])
+            assert(upper >= chunk_ranges[-1][1])
+            if lower > chunk_ranges[-1][1]:
+                chunk_ranges.append([lower, upper])
```
JBorrow, line 197: I don't really understand what's going on here - can you provide some comments?

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+            elif lower <= chunk_ranges[-1][1]:
+                chunk_ranges[-1][1] = upper
+            else:
+                raise RuntimeError("computing chunk ranges has gone horribly wrong")
+        else:
+            chunk_ranges.append([lower, upper])
+
+    return chunk_ranges
+
+@jit(forceobj=True, fastmath=True)
```
JBorrow, line 207: It's best to avoid `forceobj=True` with `numba`. `numba` as a library is only worth using with `nopython=True`.

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+def expand_ranges(ranges):
+    length = 0 
+    for bounds in ranges:
+        length += bounds[1] - bounds[0]
```
JBorrow, line 211: I don't really like this section of code... Can we replace it with something like `length = sum([x[1] - x[0] for x in ranges])`?

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+
+    output = np.zeros(length, dtype=np.int64)
+    i = 0
+    for bounds in ranges:
```
JBorrow, line 215: Yes, I can't imagine that numba will like this - it does not like iterating over 2D arrays. 
 
We should also replace this with a double for loop, so something like:

```python
index = 0

for bound_index in range(len(ranges)):
    lower = bounds[bound_index][0]
    upper = bounds[bound_index][1]
    bound_length = upper - lower

    for value in range(lower, upper):
        output[index] = value
        index += 1
```

JBorrow: If you can get away with for bound in bounds that is better but I don't think numba will like that.

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+            chunk_index += 1
+
+    # Adjust range indices
+    adjusted_ranges = ranges
+    running_sum = 0
+    for i in range(len(ranges)):
+        offset = chunks[chunk_array_index[i]][0] - running_sum
+        adjusted_ranges[i][0] = ranges[i][0] - offset
+        adjusted_ranges[i][1] = ranges[i][1] - offset
+        try:
```
JBorrow, line 243: Why do we need a try / except here?

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        upper = bounds[1]
+        bound_length = upper - lower
+        output[i:i+bound_length] = np.arange(lower, upper, dtype=np.int64)
+        i += bound_length
+
+    return output
+
+def extract_ranges_from_chunks(array, chunks, ranges):
+    # Find out which chunk range each range belongs to
+    chunk_array_index = np.zeros(len(ranges), dtype=np.int)
```
JBorrow, line 226: Probably don't need `np.int` - can we try `np.int32`?

swiftsimio/accelerated.py
```diff
@@ -168,3 +180,148 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    chunk_index = 0
+    i = 0
+    while i < len(ranges):
+        if chunks[chunk_index][0] <= ranges[i][0] and chunks[chunk_index][1] >= ranges[i][1]:
```
JBorrow, line 230: I don't really understand what's going on here - some more comments would be nice and then we can revisit.

swiftsimio/accelerated.py
aborissov: This is just to check whether the chunk we want to add to the list overlaps with the last one (to prevent e.g. [[1,10],[10, 20]] or [[1,30], [20,30]]). I think most commonly it would happen if you have two ranges within one chunk, you want to prevent adding duplicates. I'll add comments to this effect in the next commit along with the other small changes.

swiftsimio/accelerated.py
aborissov: I think that was to take care of the last element in the array. There's probably a better way to do this...

aborissov:
[subset_write_test_debug.txt](https://github.com/SWIFTSIM/swiftsimio/files/4736620/subset_write_test_debug.txt)

swiftsimio/accelerated.py
```diff
@@ -7,6 +7,8 @@
 from h5py._hl.dataset import Dataset
 
 from typing import Tuple
+from numba.typed import List
```
JBorrow, line 10: Unused import, please remove.

swiftsimio/accelerated.py
```diff
@@ -7,6 +7,8 @@
...
+from itertools import chain
```
JBorrow, line 11: Unused import, please remove.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
     ranges = ranges_from_array(mask_array)
 
     return read_ranges_from_file(handle, ranges, output_size, output_type)
+
+
+################################ ALEXEI: playing around with better read_ranges_from_file implementation #################################
```
JBorrow, line 175: Please remove this marker before merge.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+
+def concatenate_ranges(ranges: np.ndarray) -> np.ndarray:
+    """
```
JBorrow, line 178: Can we have an example added to these docs to clarify what this function does?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    ----------
+    ranges : np.ndarray
+        Array of ranges (see :func:`ranges_from_array`)
+
+    Returns
+    -------
+    np.ndarray
+        two dimensional array of ranges
+    """
+    concatenated = []
```
JBorrow, line 193: `concatenated = [ranges[0]]`

swiftsimio/accelerated.py
JBorrow: Also, please add a test for this function.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    concatenated.append(ranges[0])
+    
+    for i in range(1,len(ranges)):
```
JBorrow, line 196: If we are going to have an explicit loop here, please can you @jit the function? We might as well.

Also, this looks like it was not formatted with `black`.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        lower = ranges[i][0]
+        upper = ranges[i][1]
+        if lower <= concatenated[-1][1] + 1:
+            concatenated[-1][1] = upper
+        else:
+            concatenated.append(ranges[i])
+    
+    return np.asarray(concatenated)
```
JBorrow, line 204: Why use `asarray` here instead of just `array`?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        if len(chunk_ranges) > 0:
+            if lower > chunk_ranges[-1][1]:
+                chunk_ranges.append([lower, upper])
+            elif lower <= chunk_ranges[-1][1]:
+                chunk_ranges[-1][1] = upper
+        # If chunk_ranges is empty, don't do any checks
+        else:
+            chunk_ranges.append([lower, upper])
+
+    return np.asarray(chunk_ranges)
```
JBorrow, line 248: Again, why asarray instead of just array?

swiftsimio/accelerated.py
JBorrow: Can you replace the numpy call with this explicit second loop please?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+
+
+@jit(nopython=True, fastmath=True)
+def expand_ranges(ranges: np.ndarray) -> np.array:
```
JBorrow, line 252: Please add a unit test for this function.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        upper = ranges[i][1]
+        if lower <= concatenated[-1][1] + 1:
+            concatenated[-1][1] = upper
+        else:
+            concatenated.append(ranges[i])
+    
+    return np.asarray(concatenated)
+
+@jit(nopython=True, fastmath=True)
+def get_chunk_ranges(ranges: np.ndarray, chunk_size: np.ndarray, array_length: int) -> np.ndarray:
```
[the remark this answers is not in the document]
JBorrow: Please also add a unit test for this function. Even if it just uses mock data.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+            )
+    # Take care of the last element
+    offset = chunks[chunk_array_index[n_ranges-1]][0] - running_sum
+    adjusted_ranges[n_ranges-1][0] = ranges[n_ranges-1][0] - offset
+    adjusted_ranges[n_ranges-1][1] = ranges[n_ranges-1][1] - offset
+
+    return array[expand_ranges(adjusted_ranges)]
+
+
+def read_ranges_from_file(
```
JBorrow, line 351: Please refactor into the switched functionality that we discussed, so

```python
def read_ranges_from_file(...):
    read_ranges = read_ranges_from_file_chunked if chunked else read_ranges_from_file_unchunked

    return read_ranges(...)
```

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        upper = bounds[1]
+        bound_length = upper - lower
+        output[i : i + bound_length] = np.arange(lower, upper, dtype=np.int64)
+        i += bound_length
+
+    return output
+
+
+@jit(nopython=True, fastmath=True)
+def extract_ranges_from_chunks(
```
JBorrow, line 282: Please add a unit test for this function

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    np.ndarray
+        subset of `array` whose elements are within each range in `ranges`
+    
+    """
+    # Find out which of the chunks in the chunks array each range in ranges belongs to
+    n_ranges = len(ranges)
+    n_chunks = len(chunks)
+    chunk_array_index = np.zeros(len(ranges), dtype=np.int32)
+    chunk_index = 0
+    i = 0
```
JBorrow, line 318: Can we be clearer here that this is `range_index`?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+
+
+    # Need to get the locations of the range boundaries with
+    # respect to the indexing in the array of chunked data
+    # (as opposed to the whole dataset)
+    adjusted_ranges = np.copy(ranges)
+    running_sum = 0
+    for i in range(n_ranges-1):
+        offset = chunks[chunk_array_index[i]][0] - running_sum
+        adjusted_ranges[i][0] = ranges[i][0] - offset
```
JBorrow, line 337: Does `adjusted_ranges[i] = ranges[i] - offset` not work for both of these?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        else:
+            chunk_index += 1
+
+
+    # Need to get the locations of the range boundaries with
+    # respect to the indexing in the array of chunked data
+    # (as opposed to the whole dataset)
+    adjusted_ranges = np.copy(ranges)
+    running_sum = 0
+    for i in range(n_ranges-1):
```
JBorrow, line 335: Here, can we not say `this_chunk = chunks[chunk_array_index[i]]`, and then use this throughout?

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    Returns
+    -------
+    np.ndarray
+        subset of `array` whose elements are within each range in `ranges`
+    
+    """
+    # Find out which of the chunks in the chunks array each range in ranges belongs to
+    n_ranges = len(ranges)
+    n_chunks = len(chunks)
+    chunk_array_index = np.zeros(len(ranges), dtype=np.int32)
```
JBorrow, line 316: You just created the variable `n_ranges` for this exact purpose - use it instead of `len(ranges)` again.

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    for i in range(n_ranges-1):
+        offset = chunks[chunk_array_index[i]][0] - running_sum
+        adjusted_ranges[i][0] = ranges[i][0] - offset
+        adjusted_ranges[i][1] = ranges[i][1] - offset
+        if chunk_array_index[i + 1] > chunk_array_index[i]:
+            running_sum += (
+                chunks[chunk_array_index[i]][1] - chunks[chunk_array_index[i]][0]
+            )
+    # Take care of the last element
+    offset = chunks[chunk_array_index[n_ranges-1]][0] - running_sum
```
JBorrow, line 344: Isn't `chunk_array_index` an array of length `n_ranges`; similarly `adjusted_ranges`? In that case:

```python3
offset = chunks[chunk_array_index[-1]][0] - running_sum
adjusted_ranges[-1] = ranges[-1] - offset
```

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    """
+
+    # Get chunk size
+    if handle.chunks is not None:
+        chunk_size = handle.chunks[0]
+        
+        # Make array of chunk ranges
+        chunk_ranges = get_chunk_ranges(ranges, chunk_size, handle.shape[0])
+        chunk_range_size = int(
+            np.sum([chunk_range[1] - chunk_range[0] for chunk_range in chunk_ranges])
```
JBorrow, line 399: `chunk_ranges` is an array. We can just do:
```python

chunk_range_size = np.diff(chunk_ranges).sum()
```

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+        )
+        if isinstance(output_shape, tuple):
```
JBorrow, line 401: I would feel more comfortable with a try/except flow here:

```python
try:
    output_shape = (chunk_range_size, output_shape[1])
except TypeError:
    # Output shape is just a number, we have a 1D array.
    output_shape = chunk_range_size
```

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+            output_shape = chunk_range_size
+    else:
+        chunk_ranges = np.copy(ranges)
+
+    output = np.empty(output_shape, dtype=output_type)
+    already_read = 0
+    handle_multidim = handle.ndim > 1
+
+    for bounds in chunk_ranges:
+        read_start = bounds[0]
```
JBorrow, line 413: `read_start, read_end = bounds`

swiftsimio/accelerated.py
```diff
@@ -168,3 +170,271 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+            else np.s_[read_start:read_end]
+        )
+
+        output_dest_sel = np.s_[already_read : size_of_range + already_read]
+
+        handle.read_direct(output, source_sel=hdf5_read_sel, dest_sel=output_dest_sel)
+
+        already_read += size_of_range
+
+    chunk_length = sum([bounds[1] - bounds[0] for bounds in chunk_ranges])
```
JBorrow, line 435: Didn't we calculate this above as `chunk_range_size`?

swiftsimio/accelerated.py
```diff
@@ -168,3 +168,317 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    """
+    length = np.asarray([bounds[1] - bounds[0] for bounds in ranges]).sum()
+
+    output = np.zeros(length, dtype=np.int64)
+    i = 0
+    for bounds in ranges:
+        lower = bounds[0]
+        upper = bounds[1]
+        bound_length = upper - lower
+        output[i : i + bound_length] = np.arange(lower, upper, dtype=np.int64)
```
JBorrow, line 278: Please can you replace this line with the explicit loop that I have requested before merge? It should not be slower if we jit the function and I really do not like this version. Sorry for nit picking.

swiftsimio/accelerated.py
```diff
@@ -168,3 +168,317 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    
+    Returns
+    -------
+
+    array: np.ndarray
+        Result from reading only the relevant values from ``handle``.
+
+    See Also
+    --------
+    read_ranges_from_file_chunked: reads data within specified ranges for chunked hdf5 file
```
JBorrow, line 473: Please ensure all docstring line lengths are < 88 characters.

swiftsimio/accelerated.py
```diff
@@ -168,3 +168,317 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    read_ranges_from_file_unchunked: reads data within specified ranges for unchunked hdf5 file
+    """
+
+    average_range_size = np.diff(ranges).sum()
+    read_ranges = (
+        read_ranges_from_file_chunked
+        if handle.chunks is not None and average_range_size < 5e5
```
JBorrow, line 480: Why is this the case?

Also you have not computed an average here. You've computed a total number of particles in the variable `average_range_size`.

aborissov: In the bandwidth plots I did previously the "old" read_ranges_from_file, which is now read_ranges_from_file_unchunked, got significantly slower at small chunk sizes, but was faster for chunks greater than 1e6. So we only really want to use read_ranges_from_file_chunked when the ranges are small. 

Having said that the difference is likely to be small as the performance of the chunked version is ok even at large range sizes, so I can remove it if you prefer.

aborissov: Don't know what I was thinking when calculating the average though... Fixed it now.

JBorrow: Final thing to do before merging is to add a comment about this magic number

aborissov: Done.

swiftsimio/accelerated.py
```diff
@@ -168,3 +168,326 @@ def index_dataset(handle: Dataset, mask_array: np.array) -> np.array:
...
+    unchunked hdf5 file
+    """
+
+    # It was found that the range size for which read_ranges_from_file_chunked was
+    # faster than unchunked was approximately 5e5. For ranges larger than this the
+    # overheads associated with read_ranges_from_file_chunked caused slightly worse
+    # performance than read_ranges_from_file_unchunked
+    cross_over_range_size = 5e5
+
+    average_range_size = np.diff(ranges).sum() / len(ranges)
```
JBorrow, line 486: Sorry to be a pain but can we replace this with `np.diff(ranges).mean()`?

aborissov: No problem, sorry I'm just not as familiar as I should be about the functions provided by the libraries (even though I'm sure I have used mean and probably should have realised it exists given that sum is there). Done now.

JBorrow:
Thanks Alexei!

PR #48, opened 2020-06-09
Projection units check

aborissov:


swiftsimio/visualisation/projection.py
```diff
@@ -271,6 +271,14 @@ def project_gas_pixel_grid(
       array if you want it to be visualised the 'right way up'.
     """
 
+    # Check that the units of the rotation center correspond to those of the coordinates
+    if rotation_center is not None and (
+        isinstance(rotation_center, list)
```
JBorrow, line 276: Is this not just going to crash in the case where `rotation_center` is a list as it has no `units` attribute?

Can we do:

```python
    if rotation_center is not None:
         try:
             if rotation_center.units == data.gas.coordinates.units:
                 pass
             else:
                 raise exceptions.InvalidUnitOperation(
                      "Units of coordinates and rotation center must agree"
                 )
         except AttributeError:
              raise exceptions.InvalidUnitOperation(
                   "Ensure that rotation_center is a unyt array with the same units as coordinates"
              )
```

This check should also be applied in `project_pixel_grid` rather than `project_gas_pixel_grid` (i.e. one layer down).

aborissov: It seemed to work for me, but I agree this is more robust. Updated now.

JBorrow:
Thanks Alexei!

PR #51, opened 2020-06-11
Rotations

aborissov:


swiftsimio/visualisation/slice.py
```diff
@@ -393,11 +427,20 @@ def slice_gas(
         Data field to be projected. Default is mass. If None then simply
         count number of particles 
     
-    parallel : bool
+    parallel : bool, optional
         used to determine if we will create the image in parallel. This 
         defaults to False, but can speed up the creation of large images 
         significantly at the cost of increased memory usage.
 
+    rotation_matrix: np.array, optional
```
JBorrow, line 435: Does it provide a projection along the z axis? I don't think that's the case :).

I think we also need to add a comment to the `slice` parameter to make sure that it's clear that the z that will be used is in the rotated frame.

swiftsimio/visualisation/volume_render.py
```diff
@@ -233,6 +246,15 @@ def render_gas_voxel_grid(
         defaults to False, but can speed up the creation of large images 
         significantly at the cost of increased memory usage.
 
+    rotation_matrix: np.array, optional
+        Rotation matrix (3x3) that describes the rotation of the box around
+        ``rotation_center``. In the default case, this provides a projection
```
JBorrow, line 251: Again, same comment about `rotation_matrix` - we need a better comment describing what the rotation matrix actually does here.

tests/test_rotate_visualisations.py
```diff
@@ -0,0 +1,128 @@
...
+from swiftsimio import mask, cosmo_array, load
+from swiftsimio.visualisation.projection import project_gas
+from swiftsimio.visualisation.slice import slice_gas
+from swiftsimio.visualisation.volume_render import render_gas
+from swiftsimio.visualisation.rotation import rotation_matrix_from_vector
+import h5py
+from numpy import array_equal, mean
+
+@requires("cosmological_volume.hdf5")
+def test_rotation(filename):
```
JBorrow, line 12: Can we have a docstring here describing what's going on?

tests/test_rotate_visualisations.py
```diff
@@ -0,0 +1,128 @@
...
+    output = "single_particle.hdf5"
+    
+    data_mask = mask(filename)
```
JBorrow, line 15: Needs infromation about what's going on here - why are we doing this?

tests/test_rotate_visualisations.py
```diff
@@ -0,0 +1,128 @@
...
+    boxsize = data_mask.metadata.boxsize
+    region = [[0, b] for b in boxsize]
+    data_mask.constrain_spatial(region)
+    
+    # Write the metadata
+    infile = h5py.File(filename, "r")
+    outfile = h5py.File(output, "w")
```
JBorrow, line 22: Can we use an in-memory HDF5 file here, like we do for the other tests? There are helper functions for doing that in the test helpers.

tests/test_rotate_visualisations.py
```diff
@@ -0,0 +1,128 @@
...
+        ids.attrs.create(name, value)
+    
+    outfile.create_dataset("/PartType1/Coordinates", shape = particle_coords.shape)
+    
+    # Tidy up
+    infile.close()
+    outfile.close()
+    
+    # Start from the beginning, open the file 
+    data = load(output)
```
JBorrow, line 56: I suppose it's easier here to use this file multiple times, but it would be nice to have the differetn visualisation funtionality tested in different functions.

Can you add the code that creates a single particle file to the helper, make it in-memory, and then use this all to test in individual functions?

docs/source/visualisation/volume_render.rst
```diff
@@ -73,6 +73,58 @@ this:
    # A 1024 x 1024 x 1024 cube with dimensions of temperature
    temp_cube = mass_weighted_temp_cube / mass_cube
 
+Rotations
+---------
+
+Rotations of the box prior to slicing are provided in a similar fashion to the
```
JBorrow, line 79: Slicing -> volume rendering?

tests/helper.py
```diff
@@ -63,3 +66,66 @@ def create_in_memory_hdf5(filename="f1"):
...
+        ids.attrs.create(name, value)
+
+    # Get rid of all traces of DM
+    del outfile["/Cells/Counts/PartType1"]
+    del outfile["/Cells/Offsets/PartType1"]
+    #nparts_total = outfile["/Header"].attrs.get("NumPart_Total")
+    #nparts_this_file = outfile["/Header"].attrs.get("NumPart_ThisFile")
+    nparts_total = [2, 0, 0, 0, 0, 0]
+    nparts_this_file = [2, 0, 0, 0, 0, 0]
+    outfile["/Header"].attrs.__setitem__("NumPart_Total", nparts_total)
```
JBorrow, line 121: Why this and not just outfile["/Header"].attrs["NumPart_Total"] = nparts_total? Does that not work?

aborissov: Didn't know I could do that. Changed now.

swiftsimio/reader.py
```diff
@@ -345,6 +345,7 @@ def load_particle_types(self):
         for particle_type, particle_name in zip(
             self.present_particle_types, self.present_particle_names
         ):
+            print(particle_type, particle_name)
```
JBorrow, line 348: Please remove print statement

aborissov: ugh, don't know why I missed that. Done now

JBorrow:
Thanks Alexei!

PR #54, opened 2020-06-19
fix to_physical bug

### LINKED ISSUE #53: `to_physical()` requires data to be read in
To use the cosmology transforms we must have already accessed the data - this is unecessary and crashes in a nasty way if it is not the case.

aborissov:
Closes #53


JBorrow:
Perfect, thanks for your help on this Alexei!

PR #56, opened 2020-06-25
Read ics

### LINKED ISSUE #17: Can no longer read ICs created by swiftsimio
It appears that we can no longer read our own files. This shouldn't happen, as it should be covered by one of the unit tests.

Could be something to do with weird particle types.

aborissov:
closes #17

### DIFF
--- a/swiftsimio/writer.py
+++ b/swiftsimio/writer.py
@@ -14,7 +14,6 @@ from typing import Union, List, Callable
 from functools import reduce
 
 from swiftsimio import metadata
-from re import findall
 
 
 class __SWIFTWriterParticleDataset(object):
@@ -201,7 +200,19 @@ class __SWIFTWriterParticleDataset(object):
 
         return
 
-    def write_particle_group_metadata(self, file_handle: h5py.File, dset_attributes):
+    def write_particle_group_metadata(
+        self, file_handle: h5py.File, dset_attributes: dict
+    ):
+        """
+        Writes the relevant metadata for a particle group
+
+        Parameters
+        ----------
+        file_handle: h5py.File
+            hdf5 output file being written
+        dset_attributes: dict
+            dictionary containg metadata to attach to group
+        """
         for name, output_handle in getattr(
             metadata.required_fields, self.particle_name
         ).items():
@@ -211,7 +222,20 @@ class __SWIFTWriterParticleDataset(object):
 
         return
 
-    def get_attributes(self, scale_factor):
+    def get_attributes(self, scale_factor: float) -> dict:
+        """
+        Returns a dictionary containg the attributes to attach to the dataset
+
+        Parameters
+        ----------
+        scale_factor: float
+            the cosmological scale factor of the dataset
+
+        Returns
+        -------
+        attributes_dict: dict
+            dictionary containg the attributes applying to the dataset
+        """
         attributes_dict = {}
         a_exp_dict = {
             "coordinates": 1,
@@ -228,7 +252,7 @@ class __SWIFTWriterParticleDataset(object):
 
             # Find the scale factor associated quantities
             a_exp = a_exp_dict.get(name, 0)
-            a_factor = scale_factor * a_exp
+            a_factor = scale_factor ** a_exp
 
             attributes_dict[output_handle] = {
                 "Conversion factor to CGS (not including cosmological corrections)": [
@@ -237,13 +261,12 @@ class __SWIFTWriterParticleDataset(object):
                 "Conversion factor to physical CGS (including cosmological corrections)": [
                     field.unit_quantity.in_cgs() * a_factor
                 ],
-                "Description": b"Co-moving positions of the particles",
-                "Expression for physical CGS units": b"a U_L  [ cm ]",
-                "U_I exponent": [dim_exponents[0]],
+                # Assign the exponents in the proper order (see unyt.dimensions.base_dimensions)
+                "U_I exponent": [dim_exponents[5]],
                 "U_L exponent": [dim_exponents[1]],
-                "U_M exponent": [dim_exponents[2]],
+                "U_M exponent": [dim_exponents[0]],
                 "U_T exponent": [dim_exponents[3]],
-                "U_t exponent": [dim_exponents[4]],
+                "U_t exponent": [dim_exponents[2]],
                 "a-scale exponent": [a_exp],
                 "h-scale exponent": [0.0],
             }
@@ -251,12 +274,39 @@ class __SWIFTWriterParticleDataset(object):
         return attributes_dict
 
 
-def get_dimensions(obj):
-    dimensions = ["current", "length", "mass", "temperature", "time"]
-    n_dims = len(dimensions)
-    exp_array = np.zeros(n_dims, dtype=np.float32)
-    dim_array = [x.as_base_exp() for x in obj.as_ordered_factors()]
+def get_dimensions(dimension: unyt.dimensions) -> np.ndarray:
+    """
+    Returns exponents corresponding to base dimensions for given unyt dimensions object
 
+    Parameters
+    ----------
+    dimension: unyt.dimensions
+        dimension for which we're identifying the exponents
+
+    Returns
+    -------
+    exp_array: np.ndarray
+        array of exponents corresponding to each base dimension
+
+    Examples
+    --------
+    >>> get_dimensions(unyt.dimensions.velocity)
+    np.array([0,1,-1,0,0,0,0,0,0])
+
+    """
+    # Get the names of all the dimensions
+    dimensions = [str(x) for x in unyt.dimensions.base_dimensions]
+    # Annoyingly it's current_mks instead of current in unyt, so change that
+    dimensions[5] = "(current)"
+    n_dims = len(dimensions)
+
+    # create the return array
+    exp_array = np.zeros(n_dims, dtype=np.float32)
+
+    # extract the base and exponent for each of the units
+    dim_array = [x.as_base_exp() for x in dimension.as_ordered_factors()]
+
+    # Find out which dimensions and exponents are present in the base units
     for i in range(n_dims):
         for dim in dim_array:
             if dimensions[i] in str(dim[0]):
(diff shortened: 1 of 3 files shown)


swiftsimio/writer.py
```diff
@@ -14,6 +14,7 @@
 from functools import reduce
 
 from swiftsimio import metadata
+from re import findall
```
JBorrow, line 17: Unused import

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
 
         return
 
+    def write_particle_group_metadata(self, file_handle: h5py.File, dset_attributes):
```
JBorrow, line 204: What are dset attributes? What type does it have?

Can we have a docstring here please.

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+        for name, output_handle in getattr(
+            metadata.required_fields, self.particle_name
+        ).items():
+            obj = file_handle[f"/PartType{self.particle_type}/{output_handle}"]
+            for attr_name, attr_value in dset_attributes[output_handle].items():
+                obj.attrs.create(attr_name, attr_value)
+
+        return
+
+    def get_attributes(self, scale_factor):
```
JBorrow, line 214: Docstring here again, as well as documentation.

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+            metadata.required_fields, self.particle_name
+        ).items():
+            field = getattr(self, name)
+
+            # Find the exponents for each of the dimensions
+            dim_exponents = get_dimensions(field.units.dimensions)
+
+            # Find the scale factor associated quantities
+            a_exp = a_exp_dict.get(name, 0)
+            a_factor = scale_factor * a_exp
```
JBorrow, line 231: `a_factor = scale_factor ** a_exp`?

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+                "U_T exponent": [dim_exponents[3]],
+                "U_t exponent": [dim_exponents[4]],
+                "a-scale exponent": [a_exp],
+                "h-scale exponent": [0.0],
+            }
+
+        return attributes_dict
+
+
+def get_dimensions(obj):
```
JBorrow, line 254: Please add documentation here.

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+    dimensions = ["current", "length", "mass", "temperature", "time"]
```
JBorrow, line 255: Is this list not present anywhere in unyt?

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+
+            attributes_dict[output_handle] = {
+                "Conversion factor to CGS (not including cosmological corrections)": [
+                    field.unit_quantity.in_cgs()
+                ],
+                "Conversion factor to physical CGS (including cosmological corrections)": [
+                    field.unit_quantity.in_cgs() * a_factor
+                ],
+                "Description": b"Co-moving positions of the particles",
+                "Expression for physical CGS units": b"a U_L  [ cm ]",
```
JBorrow, line 241: `b"a U_L [cm]"` is not correct for all items. I think we should just leave this as a blank string.

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+            a_factor = scale_factor * a_exp
+
+            attributes_dict[output_handle] = {
+                "Conversion factor to CGS (not including cosmological corrections)": [
+                    field.unit_quantity.in_cgs()
+                ],
+                "Conversion factor to physical CGS (including cosmological corrections)": [
+                    field.unit_quantity.in_cgs() * a_factor
+                ],
+                "Description": b"Co-moving positions of the particles",
```
JBorrow, line 240: Not everything that this will write will be a comoving position.

swiftsimio/writer.py
```diff
@@ -200,6 +201,69 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+
+        return attributes_dict
+
+
+def get_dimensions(obj):
+    dimensions = ["current", "length", "mass", "temperature", "time"]
+    n_dims = len(dimensions)
+    exp_array = np.zeros(n_dims, dtype=np.float32)
+    dim_array = [x.as_base_exp() for x in obj.as_ordered_factors()]
+
```
JBorrow, line 259: This piece of code is quite confusing, I don't fully understand what the stuff here is.

I suppose `dim_array` is an array of dimensions? And `exp_array` is the exponents corresponding to that?

Or is `dim_array` of shape (n_dims, 2) with it being a list of lists of name, power?

tests/test_read_ic.py
```diff
@@ -0,0 +1,48 @@
...
+    x.write("test.hdf5")
+
+    data = load("test.hdf5")
+
+    assert np.array_equal(data.gas.coordinates, x.gas.coordinates)
+    assert np.array_equal(data.gas.velocities, x.gas.velocities)
+    assert np.array_equal(data.gas.masses, x.gas.masses)
+    assert np.array_equal(data.gas.internal_energy, x.gas.internal_energy)
+    assert np.array_equal(data.gas.smoothing_length, x.gas.smoothing_length)
+
```
JBorrow, line 47: Tidy up `test.hdf5`.

Also we should give this a better name than `test.hdf5`, maybe `test_write_output_units.hdf5`

tests/test_read_ic.py
```diff
@@ -0,0 +1,48 @@
+from swiftsimio import load
+from swiftsimio import Writer
+from swiftsimio.units import cosmo_units
+
+import unyt
+import numpy as np
+
+
+def test_read():
```
JBorrow, line 9: A better name than `test_read` would be nice.

swiftsimio/writer.py
aborissov: removed

swiftsimio/writer.py
aborissov: done

swiftsimio/writer.py
aborissov: done

swiftsimio/writer.py
aborissov: really don't know what happened here... fixed now.

swiftsimio/writer.py
aborissov: done

swiftsimio/writer.py
aborissov: There is indeed. I changed it to use the one provided in unyt. unfortunately a bit of hacking needed to happen because they use current_mks instead of current.

swiftsimio/writer.py
aborissov: ugh, I thought I had removed these... gone now.

swiftsimio/writer.py
aborissov: Yes, it is a list of lists of unyt.dimensions, powers. I've added a comment to this effect.

tests/test_read_ic.py
aborissov: tidied up.

tests/test_read_ic.py
aborissov: I hate naming things, hopefully test_reading_ic_units is better.

swiftsimio/writer.py
```diff
@@ -200,6 +200,120 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+        file_handle: h5py.File
+            hdf5 output file being written
+        dset_attributes: dict
+            dictionary containg metadata to attach to group
+        """
+        for name, output_handle in getattr(
+            metadata.required_fields, self.particle_name
+        ).items():
+            obj = file_handle[f"/PartType{self.particle_type}/{output_handle}"]
+            for attr_name, attr_value in dset_attributes[output_handle].items():
```
JBorrow, line 220: Maybe I am wrong, but can we not just do:

```
obj.attrs = dset_attributes[output_handle]
```

swiftsimio/writer.py
```diff
@@ -200,6 +200,120 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+        scale_factor: float
+            the cosmological scale factor of the dataset
+
+        Returns
+        -------
+        attributes_dict: dict
+            dictionary containg the attributes applying to the dataset
+        """
+        attributes_dict = {}
+        a_exp_dict = {
```
JBorrow, line 240: I would prefer it if we could move this to the `metadata`. We already have https://github.com/SWIFTSIM/swiftsimio/blob/master/swiftsimio/metadata/cosmology/cosmology_fields.py, maybe you can figure out how to integrate that here? I don't think we use `cosmology_fields.py` any more, so it may be worth just editing that to make it easier for you.

swiftsimio/writer.py
```diff
@@ -200,6 +200,120 @@ def write_particle_group(self, file_handle: h5py.File, compress: bool):
...
+                "U_T exponent": [dim_exponents[3]],
+                "U_t exponent": [dim_exponents[2]],
+                "a-scale exponent": [a_exp],
+                "h-scale exponent": [0.0],
+            }
+
+        return attributes_dict
+
+
+def get_dimensions(dimension: unyt.dimensions) -> np.ndarray:
```
JBorrow, line 277: I don't like this returning an array like this - can we make this a dictionary, or an object that it returns? Seeing as this is immutable maybe it should return a `namedtuple`.

swiftsimio/writer.py
aborissov: Unfortunately not. it claims it can't set the attribute that way.

JBorrow: :( Boo. Thanks for checking.

swiftsimio/writer.py
aborissov: Changed this to a dictionary.

swiftsimio/writer.py
aborissov: done.

JBorrow:
Thanks Alexei!

PR #63, opened 2020-06-26
check subset write test on subset of subset of snapshot

aborissov:


JBorrow:
Looks good to me, if this works let's merge it.

aborissov:
It worked when I tested it, so just waiting for the server

PR #66, opened 2020-07-27
WIP: Iterative IC generation

mladenivkovic:
Heya

Here's a first working version of the IC making algorithm. I extended it to work in 1D, 2D, 3D, and with arbitrary boxsizes.
It's at a reviewable state, so please have a look at it when you get the time, and let me know what you think and what needs to be improved.
I haven't finished the documentation yet, but I left a script in a new directory `/tmp/`, where you can run it if you want to.

Todo's left:
- finish testing
- remove printing and plotting stuff (still needed for testing)
- proper documentation (in progress)
- unit tests
- remove /tmp/ directory when we don't need it any longer


swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+    W.gas.velocities = unyt_array(np.zeros(npart, dtype=np.float), unyt.m / unyt.s)
+
+    # If IDs are not present, this automatically generates
+    fname = "IC-generation-iteration-" + str(iteration).zfill(5) + ".hdf5"
+    W.write(fname)
+
+    return
+
+
+def _IC_plot_current_situation(
```
JBorrow, line 1002: We should at the very least move this to a separate file. 

I don't like having underscores here, that implies they are private and unsed internally rather than for debugging. We can create `swiftsimio/initial_conditions/debug.py` and put this there. Or `swiftsimio/initial_conditions/plotting.py`.

mladenivkovic:
When you say 'move to an object-oriented approach here', what exactly do you have in mind? To put every method here inside one class?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+from typing import Union
+
+from .IC_kernel import get_kernel_data
+from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
+
+
+seed_set = False
+
+
+def IC_set_IC_params(
```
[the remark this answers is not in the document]
mladenivkovic: Seems like a good idea indeed. Should I keep the setter method nonetheless?

JBorrow:
So what I would have done here is created a class, say:

```python
class ParticleGenerator(object):
    # All of the default properties
    boxsize: ...
    ...
    def __init__(self, rho: callable):
        # Basically does nothing because we have already set all of the
        # default properties above

        self.rho = rho

        return

    def initial_setup(self, ...):
         # Initial set-up of the first grid (i think that's how this works).

         # This can call out to the two functions, the one that creates the uniform grid
         # and the one that does the rejection sampling, whichever is appropriate.
         return


    def perform_iteration(self, ...)
         # Does one iteration of the algorithm to move the particles a bit closer to ideal rho


    def iterate(self, ...):
         # Iterates to convergence. Can also take an optional RunParameters instance that
         # can store the iteration requirements?
```

You can then have your debug functions take an instance of this, or provide things like `stats` in postprocessing as a property.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
JBorrow: Setter method?

swiftsimio/initial_conditions/IC_kernel.py
```diff
@@ -0,0 +1,186 @@
...
+    Note
+    ------------------
+    + The return value is not normalized. It needs to be divided by ``H^ndim`` 
+      and multiplied by the kernel norm to get the proper values. 
+    """
+
+    q = r / H
+    W = np.zeros(q.shape)
+
+    #  if q <= 1.:
```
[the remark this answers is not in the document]
mladenivkovic: Has been changed according to your suggestion.

swiftsimio/initial_conditions/IC_kernel.py
```diff
@@ -0,0 +1,186 @@
...
+
+        - H:  np.ndarray or float
+          compact support radius of kernel. Scalar or numpy array of same 
+          shape as ``r``.
+
+    kernel_gamma: float
+        H/h (compact support radius / smoothing length) for given kernel and 
+        dimension
+    
+    """
```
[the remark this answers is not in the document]
mladenivkovic: Has been changed according to your first suggestion.
We can still add the second later, it's not much work, but let's first decide how we're going to handle all kernels in the package and then do more changes.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Nevermind, old java lingo ^^

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+        understands them.
+
+
+    Notes
+    -----------
+    
+    + The returned dict is a required argument to call ``generate_IC_for_given_density()``
+    """
+
+    icSimParams = {}
```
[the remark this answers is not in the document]
mladenivkovic: Has been changed according to your suggestion.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+        icSimParams["unit_l"] = unit_l
+        icSimParams["boxsizeToUse"] = boxsize.to(unit_l).value
+    icSimParams["unit_m"] = unit_m
+    icSimParams["kernel"] = kernel
+    icSimParams["eta"] = eta
+
+    return icSimParams
+
+
+def IC_set_run_params(
```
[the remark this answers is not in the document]
mladenivkovic: Has been changed according to your suggestions.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+    icRunParams["DELTA_INIT"] = delta_init
+    icRunParams["DELTA_REDUCT"] = delta_reduction_factor
+    icRunParams["DELTA_MIN"] = delta_min
+    icRunParams["REDIST_FREQ"] = redistribute_frequency
+    icRunParams["REDIST_FRAC"] = redistribute_fraction
+    icRunParams["REDIST_STOP"] = no_redistribution_after
+    icRunParams["REDIST_REDUCT"] = redistribute_fraction_reduction
+    icRunParams["DUMPFREQ"] = intermediate_dump_frequency
+
+    global seed_set
```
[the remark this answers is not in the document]
mladenivkovic: In case somebody (like me) has the very bright idea to call the setter routine twice. In that case, the seed gets reset, and that's most likely not something you want. Example:

```
>>> import numpy as np
>>> np.random.seed(666)
>>> np.random.uniform(size=10)
array([0.70043712, 0.84418664, 0.67651434, 0.72785806, 0.95145796,
       0.0127032 , 0.4135877 , 0.04881279, 0.09992856, 0.50806631])
>>> np.random.seed(666)
>>> np.random.uniform(size=10)
array([0.70043712, 0.84418664, 0.67651434, 0.72785806, 0.95145796,
       0.0127032 , 0.4135877 , 0.04881279, 0.09992856, 0.50806631])

```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+
+seed_set = False
+
+
+def IC_set_IC_params(
+    boxsize: unyt_array = unyt_array([1.0, 1.0, 1.0], "cm"),
+    periodic: bool = True,
+    nx: int = 100,
+    ndim: int = 2,
+    unit_l: Union[None, unyt.unit_object.Unit, str] = None,
```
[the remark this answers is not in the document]
mladenivkovic: The ``None`` is an option here because if it's left out, I take the same units as is given for boxsize, which in turn is required to be a unyt_array.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+    Returns
+    ------------------
+
+    x: unyt_array 
+        unyt_array(shape=(npart, 3), dtype=float) of coordinates, where 
+        ``npart = nx ** ndim``, both of which are set in the ``icSimParams``
+        dict
+    """
+
+    nx = icSimParams["nx"]
```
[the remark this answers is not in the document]
mladenivkovic: I'd prefer to keep it as to not have to deal with unit conversions explicitly inside the function again. I was planning on doing that only once when the ic_sim_params object is instantiated, and let the user not thing about that any more.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+    periodic = icSimParams["periodic"]
+    npart = nx ** ndim
+    maxdelta = (
+        0.4 * np.mean(boxsize) / nx
+    )  # maximal deviation from uniform grid of any particle along an axis
+
+    x = IC_uniform_coordinates(icSimParams)  # generate uniform grid first
+
+    for d in range(ndim):
+        sign = 2 * np.random.randint(0, 2, size=npart) - 1  # get +1 or -1
```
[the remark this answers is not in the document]
mladenivkovic: because old habits die hard. Fixed now.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+                start = j * nx ** 2 + i * nx
+                stop = j * nx ** 2 + (i + 1) * nx
+                x[start:stop, 0] = xcoords
+                x[start:stop, 1] = ycoords[i]
+                x[start:stop, 2] = zcoords[j]
+
+    return x
+
+
+def IC_perturbed_coordinates(icSimParams: dict):
```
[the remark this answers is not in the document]
mladenivkovic: Main idea of the `IC_perturbed_coordinates` function was to give the user the option to quickly create this kind of coordinates to give the iterating method a first guess. (The `IC_uniform_coordinates` function is used in other places in the module.) That mindset doesn't really fit in with making it an object method in my eyes.

I've no objections to more fun names, just suggest whatever you'd like.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+        raise TypeError("rho_anal needs to return a numpy array as the result.")
+
+    # shortcuts and constants
+    nx = icSimParams["nx"]
+    ndim = icSimParams["ndim"]
+    periodic = icSimParams["periodic"]
+    boxsize = icSimParams["boxsizeToUse"]
+    npart = nx ** ndim
+
+    MID = 1.0
```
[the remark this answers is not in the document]
mladenivkovic: Has been changed according to your suggestion.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+        )
+
+    try:
+        res = rho_anal(np.ones((10, 3), dtype="float"), icSimParams["ndim"])
+    except TypeError:
+        errmsg = (
+            "rho_anal must take only a numpy array x of coordinates as an argument."
+        )
+        raise TypeError(errmsg)
+    if type(res) is not np.ndarray:
```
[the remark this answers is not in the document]
mladenivkovic: Fixed.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+            xr[0, d] = np.random.uniform(low=0.0, high=coord_threshold[d])
+
+        if np.random.uniform() <= rho_anal(xr, ndim) / rho_max:
+            x[keep] = xr
+            keep += 1
+
+    return unyt_array(x, icSimParams["unit_l"])
+
+
+def generate_IC_for_given_density(
```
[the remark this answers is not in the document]
mladenivkovic: I'm not convinced that's a great idea. I intended this function to be the only one you'd have to call to get your initial conditions, and to have to deal with the simulation parameters explicitly before that by calling the appropriate functions that produce the dicts (which are classes now). Essentially, you can get your ICs in 3-4 lines right now. What's the advantage of putting it into a class like you're suggesting?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+                iteration, min_deviation, av_deviation, max_deviation
+            )
+        )
+
+        if (
+            max_deviation < icRunParams["DISPL_THRESH"]
+        ):  # don't think about stopping until max < threshold
+            unconverged = dev[dev > icRunParams["CONVERGE_THRESH"]].shape[0]
+            if unconverged < icRunParams["TOLERANCE_PART"] * npart:
+                print("Convergence criteria are met.")
```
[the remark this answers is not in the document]
mladenivkovic: As I wrote in the descriptions, those are still here for testing purposes. Everything containing printing and plotting inside this module is not meant to stay.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+        if touched[oind]:
+            continue  # skip touched particles
+
+        # do we work with it?
+        othresh = (rho[oind] - rhoA[oind]) / rho[oind]
+        othresh = erf(othresh)
+        if np.random.uniform() < othresh:
+
+            attempts = 0
+            while True:
```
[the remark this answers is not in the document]
mladenivkovic: Fixed

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+
+
+    Returns
+    -------------------
+
+    Nothing.
+
+    """
+
+    from swiftsimio import Writer
```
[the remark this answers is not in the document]
mladenivkovic: Fixed

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+
+    nx = icSimParams["nx"]
+    ndim = icSimParams["ndim"]
+    npart = nx ** ndim
+
+    ICunits = unyt.UnitSystem(
```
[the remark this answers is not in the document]
mladenivkovic: I didn't think it was necessary. The only place where I demand units is in the ``boxsize`` (I added an explicit check now) parameter, the rest is voluntary. And if ``unit_l is None``, then take the same units as ``boxsize``.  The only reason I have a ``UnitSystem`` here is because from what I saw, ``Writer`` needs one.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1281 @@
...
+    )
+
+    # invent some junk to fill up necessary arrays
+    W.gas.internal_energy = unyt_array(
+        np.zeros(npart, dtype=np.float), unyt.m ** 2 / unyt.s ** 2
+    )
+    W.gas.velocities = unyt_array(np.zeros(npart, dtype=np.float), unyt.m / unyt.s)
+
+    # If IDs are not present, this automatically generates
+    fname = "IC-generation-iteration-" + str(iteration).zfill(5) + ".hdf5"
```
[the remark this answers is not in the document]
mladenivkovic: Fixed.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: I meant debugging _before we merge this_. I initially intended to remove it fully at the end, but with the intermediate data dumps, i could just use it for a script/example to quickly just plot these or something. Or we might scratch it anyway. Anyhow, I moved it to its own file now.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+
+from .IC_kernel import get_kernel_data
+from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
+from swiftsimio import Writer
+
+
+seed_set = False
+
+
+class ic_sim_params(object):
```
JBorrow, line 25: Sorry, I wasn't clear earlier. This should be `ICSimParams` - we do use uppercase for class names.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+    r"""
+    object containing simulation parameters
+    """
+    boxsize: unyt_array = unyt_array([1.0, 1.0, 1.0], "cm")
+    periodic: bool = True
+    nx: int = 100
+    ndim: int = 2
+    unit_l: Union[None, unyt.unit_object.Unit, str] = None
```
JBorrow, line 33: I really think it would be best to replace this with an `unyt.unit_system`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+    unit_m: Union[unyt.unit_object.Unit, str] = "g"
+    kernel: str = "cubic spline"
+    eta: float = 1.2348
+
+    def __init__(
+        self,
+        boxsize: unyt_array = unyt_array([1.0, 1.0, 1.0], "cm"),
```
JBorrow, line 40: You don't need to initialise these here - you can just assume they are set from the above. If you want to change them, you can just do:

```python
ic_params = ICSimParams()
ic_params.boxsize = unyt.unyt_array([1.0, 0.4, 0.2], "cm")
```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+        kernel: str = "cubic spline",
+        eta: float = 1.2348,
+    ):
+        r"""
+        Set up the simulation parameters for the initial conditions you want to
+        generate. The dict it returns is a necessary argument to call
+        ``generate_IC_for_given_density()``
+
+
+        Parameters
```
JBorrow, line 55: You can then move this up to the top-level docsrting under `attributes`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+            self.boxsize_to_use = boxsize.value  # use only np.array
+        else:
+            self.unit_l = unit_l
+            self.boxsize_to_use = boxsize.to(unit_l).value
+        self.unit_m = unit_m
+        self.kernel = kernel
+        self.eta = eta
+        return
+
+    def copy(self):
```
JBorrow, line 118: This is not necessary.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1047 @@
...
+    delta_reduct: float = 1.0
+    delta_min: float = 1e-6
+    redist_freq: int = 20
+    redist_frac: float = 0.01
+    redist_reduct: float = 1.0
+    no_redistribution_after: int = 200
+    dump_basename: str = "IC-generation-iteration-"
+    dumpfreq: int = 50
+
+    def __init__(
```
JBorrow, line 153: same comments as above for your `__init__`.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
JBorrow: I'm not necessarily concerned about the number of lines that it takes to produce an initial condition; my thinking is more that if we can have a better abstraction that is easier for people to understand (and that is consistent with the rest of `swiftsimio`) that's going to be better.

You can take, for instance, the `Writer` class. This is a bit of a monolith; it performs unit conversion, writes to HDF5 files, deals with metadata, and can even generate data if required.

Even though it does all of these jobs, it has a singular purpose - you have a set of data, and you would like a valid SWIFT HDF5 file. 

I think the equivalent for this system is that you have a function describing the density distribution and you would like a set of particle positions that describe that function.

To fit with the other abstraction, then, we want a singular object (called `ParticleGenerator` or something) that takes this function as an argument, and after we have called some methods makes available a set of co-ordinates as a property. So the workflow would go something like this:

```python
# Take the numbers that 'pass' to the outputs, like boxsize and number of particles
# as arguments
particles = ParticleGenerator(
    rho=my_density_function,
    number_of_particles=128,
    number_of_dimensions=3,
    boxsize=unyt.unyt_array([1.0, 0.2, 0.3], "kpc"),
    unit_system=unyt.unit_system("cgs"),
)
# Modify 'internal' parameters, such as convergence criteria by changing
# attributes
particles.some_config_parameter = 3
particles.some_other_config_parameter = "hello"
# User will need to make a choice of how to set the initial setup
particles.initial_setup(method="rejection")
# Now get the converged answer
particles.iterate()
# Result is available as a property
print(particles.positions)
```

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
JBorrow: To make this consistent with the rest of `swiftsimio` it would be best to take a `UnitSystem` at the top level.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Fixed

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Has been removed in the updated version.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Fine, have it your way.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Fixed.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: It takes a ``unyt.unit_system.UnitSystem`` as the argument now.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: 'tis done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: not applicable any longer.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: 'tis done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: 'tis done.

mladenivkovic:
Updated according to your advice, docstrings have been cleaned up. Let me know what you think. Meanwhile, I'll be writing a documentation and some examples.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+from swiftsimio import Writer
+
+
+class RunParams(object):
+    r"""
+    A class to store iteration runtime parameters. Before running the iteration,
+    these parameters are intended to be direcly accessed by the user.
+
+
+    Attributes
```
JBorrow, line 23: Styling request: please style this so that there are the same number of dashes as characters, and have a space between the items and the attributes so:

```
Attributes
----------

iter_max: int
```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    ---------------
+    self.iter_max: int
```
JBorrow, line 25: You don't need `self.` here, just `iter_max`.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        maximal numbers of iterations to run for generating IC conditions
+
+    self.iter_min: int
+        minimal numbers of iterations to run for generating IC conditions
+
+    self.converge_thresh: float
```
JBorrow, line 31: Can we just have the full names here? So e.g.

+ `iter_max` -> `max_iterations`
+ `converge_thresh` -> `convergence_threshold`
+ `tolerance_frac` -> `tolerance_fraction`
+ `displ_thresh` -> `displacement_threshold`
+ `delta_reduct` -> `delta_reduction`
+ ...

It helps make the code much more readable.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    displ_thresh: float = 1e-3
+    delta_init: Union[float, None] = None
+    delta_reduct: float = 1.0
+    delta_min: float = 1e-6
+    redist_freq: int = 20
+    redist_frac: float = 0.01
+    redist_reduct: float = 1.0
+    redist_stop: int = 200
+    dump_basename: str = "IC-generation-iteration-"
+    dumpfreq: int = 50
```
JBorrow, line 113: `dumpfreq` -> `dump_frequency`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    tolerance_part: float = 1e-3
+    displ_thresh: float = 1e-3
+    delta_init: Union[float, None] = None
+    delta_reduct: float = 1.0
+    delta_min: float = 1e-6
+    redist_freq: int = 20
+    redist_frac: float = 0.01
+    redist_reduct: float = 1.0
+    redist_stop: int = 200
+    dump_basename: str = "IC-generation-iteration-"
```
JBorrow, line 112: Default `dump_basename` should not have dashes in it - `IC_generation_iteration`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    dumpfreq: int = 50
+
+    def __init__(self):
+        # always set one random seed.
+        self.set_random_seed(666)
```
JBorrow, line 117: Should we not allow the seed to be random in the base case?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        return
+
+    def set_random_seed(self, seed: int = 666):
```
JBorrow, line 120: Let's make `seed` a required parameter here.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        norm to be used to compute the displacement "force"
+
+    self.delta_r_norm_min: float
+        lower threshold for "force" normalisation.
+
+    self.compute_delta_r_norm: bool
+        whether the user wants the code to compute an appropriate
+        normalisation.
+
+    self.Nngb: float
```
JBorrow, line 156: This should be `neighbours` to be consistent with the rest of `swiftsimio`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        number of neighbours to be searched for
+
+    self.Nngb_int: int
```
JBorrow, line 159: Let's just round `self.Nngb` up for this.

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    max_displacement: np.ndarray
+        stores the maximal displacement of every iteration step.
+
+    min_displacement: np.ndarray
+        stores the minimal displacement of every iteration step.
+
+    avg_displacement: np.ndarray
+        stores the average displacement of every iteration step.
+
+    niter: int
```
JBorrow, line 197: `number_of_iterations`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        total number of particles in simulation
+
+    self.coordinates: unyt.unyt_array
+        unyt_array containing the final particle coordinates with shape (self.npart, 3).
+
+    self.masses: unyt.unyt_array
+        unyt_array containing the final particle masses with shape (self.npart).
+
+    self.smoothing_lengths: unyt.unyt_array
+        unyt_array containing the final particle smoothing lengths with shape (self.npart, 3).
```
JBorrow, line 322: Make sure that your comments don't exceed 88 characters as unfortunately `black` doesn't wrap them :(

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        unitless particle masses that will be worked with/on with shape (self.npart).
+
+
+
+
+    Notes
+    -----------
+    
+    + The default workflow should be something like
+        ``pg = ParticleGenerator(...)``
```
JBorrow, line 340: You should use a `code-block` here like:

```
Notes
-----

The default workflow looks something like:

.. code-block:: python

    pg = ParticleGenerator()

    pg.runparams.iter_max = 10000
    pg.runparams.whatever_parameter = ...

    pg.initial_setup()

    pg.run_iteration()

File: swiftsimio/initial_conditions/generate_particles.py [hunk 38]
```diff
@@ -1,5 +1,1304 @@
...
+
+        ``pg.run_iteration()``
+
+        ``print("profit")``
+    """
+
+    # simulation parameters
+    rhofunc: callable
+    boxsize: unyt.unyt_array
+    unitsys: unyt.unit_systems.UnitSystem
```
JBorrow, line 356: `unit_system`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    nx: int
```
JBorrow, line 357: `number_of_particles`?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+
+        ``pg.initial_setup(...)``
+
+        ``pg.run_iteration()``
+
+        ``print("profit")``
+    """
+
+    # simulation parameters
+    rhofunc: callable
```
JBorrow, line 354: `density_function` or `rho_function` or `rho`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+    boxsize: unyt.unyt_array
+    unitsys: unyt.unit_systems.UnitSystem
+    nx: int
+    ndim: int
+    periodic: bool
+    kernel: str
+    eta: float
+    rho_max: Union[float, None]
+
+    runparams: RunParams
```
JBorrow, line 364: `run_params` and `iter_params`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        npart = self.npart
+        rhofunc = self.rhofunc
+        x = self.x
+        runparams = self.runparams
+        ipars = self.iterparams
+
+        from .IC_plotting import IC_plot_current_situation
+
+        # kernel data
+        kernel_func, _, kernel_gamma = get_kernel_data(self.kernel, ndim)
```
JBorrow, line 919: Unused parameter `kernel_gamma`, is that intended?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+                boundary = boxsize[d]
+                x[x > boundary] -= delta_r[x > boundary]
+                x[x < 0.0] -= delta_r[x < 0.0]
+
+        # reduce delta_r_norm
+        ipars.delta_r_norm *= runparams.delta_reduct
+        # assert minimal delta_r
+        ipars.delta_r_norm = max(ipars.delta_r_norm, ipars.delta_r_norm_min)
+
+        # get displacements in units of mean interparticle distance
```
JBorrow, line 1039: Here can we not just do:

```
displacement = np.sqrt(np.sum(delta_r * delta_r, axis=1))
```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        displacement = np.zeros(npart, dtype=np.float)
+        for d in range(ndim):
+            displacement += delta_r[:, d] ** 2
+        displacement = np.sqrt(displacement)
+
+        max_displacement = displacement.max()
+        max_displacement /= ipars.mean_interparticle_distance
+        min_displacement = displacement.min()
+        min_displacement /= ipars.mean_interparticle_distance
+        avg_displacement = displacement.sum() / displacement.shape[0]
```
JBorrow, line 1049: `avg_displacement = displacement.mean()`?

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+            ``None``, no particles have been moved.
+        """
+
+        x = self.x
+        npart = x.shape[0]
+        boxsize = self.boxsize_to_use
+        ndim = self.ndim
+
+        # how many particles are we moving?
+        to_move = int(npart * self.runparams.redist_frac + 0.5)
```
JBorrow, line 1162: This can be simplified:

```
to_move = int(npart * self.runparams.redist_frac + 0.5)

if to_move <= 0:
  return None
```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+        W.gas.masses = unyt.unyt_array(m, u_m)
+        W.gas.densities = unyt.unyt_array(rho, u_m / u_l ** ndim)
+
+        # invent some junk to fill up necessary arrays
+        W.gas.internal_energy = unyt.unyt_array(
+            np.ones(npart, dtype=np.float), u_l ** 2 / u_t ** 2
+        )
+        W.gas.velocities = unyt.unyt_array(np.zeros(npart, dtype=np.float), u_l / u_t)
+
+        fname = self.runparams.dump_basename + str(iteration).zfill(5) + ".hdf5"
```
JBorrow, line 1301: I would prefer if we could use f-string formatting here and 4-filled zeros to be consistent with the rest of `swiftsimio` and SWIFT itself:

```
filename = f"{self.runparams.dump_basename}_{iteration:04d}.hdf5"
```

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1,5 +1,1304 @@
...
+import numpy as np
+import unyt
+from math import erf
+from typing import Union
+import warnings
+
+from .IC_kernel import get_kernel_data
+from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
+from swiftsimio import Writer
+
```
JBorrow, line 15: This is a massive improvement and I really like the new object-oriented design. I now, unfortunately, have some nit-picky comments :p.

Thanks again for all of your hard work on this!

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: PyCharm was complaining about not having the `self` though

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: I disagree. I think in this case, where you might have to start over multiple times and tweak your runtime parameters a bit to get what you want, it is crucial to get consistent results with every rerun.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Is that really that necessary? As far as I see, sphinx doesn't really care about where I put linebreaks

swiftsimio/initial_conditions/generate_particles.py
JBorrow: The problem is that in our ReadTheDocs output it ends up looking like this: 
image
which just has unecessary self's.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: I wouldn't. I had plenty of cases where I ended up working with 10000+ iterations.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Not applicable any longer.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Not applicable any longer.

swiftsimio/initial_conditions/generate_particles.py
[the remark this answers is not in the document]
mladenivkovic: Optional arguments for ``nx`` and ``ndim`` have been added.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Yes, just to mess with you.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Changed to f-string formatting. Keeping the 5 digits.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Done.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Did it anyway.

mladenivkovic:
Finally I added a documentation and unit tests. Once you're happy with the contents, I can remove all the printing and plotting and be finally done with this.

JBorrow:
I will need this to pass the test suite before reviewing

mladenivkovic:
The ``ParticleGenerator()`` now has an internal random number generator that doesn't affect the global ``numpy`` one outside of it.

swiftsimio/initial_conditions/generate_particles.py
JBorrow: I still don't fully agree with this. Every other python library uses a random initial seed by default (numpy, scipy, etc.), and we should too. I would like to see this initial random seed before I merge.

JBorrow:
I currently get a crash with the following input:

```
"""
Makes Sod Shock initial conditions using the particle generator.
"""

import numpy as np
import unyt

from swiftsimio.initial_conditions import ParticleGenerator

def rho(x, ndim):
    """
    Density field in the initial SodShock.
    """

    output = np.ones(len(x), dtype=np.float32)
    output[x[:, 0] > 1.0] = 0.125

    return output

boxsize = unyt.unyt_array([2.0, 1.0, 1.0], "cm")
unit_system = unyt.unit_systems.cgs_unit_system
number_of_particles = 32
ndim = 3

generator = ParticleGenerator(
    rho,
    boxsize,
    unit_system,
    number_of_particles,
    ndim
)

generator.initial_setup()
generator.run_iteration()
```

What am I doing wrong?

Error:
```
  File "make_sod_ics.py", line 34, in <module>
    generator.run_iteration()
  File "/Users/mphf18/Documents/GitHub/swiftsimio/swiftsimio/initial_conditions/generate_particles.py", line 1359, in run_iteration
    self.iteration_step(iteration)
  File "/Users/mphf18/Documents/GitHub/swiftsimio/swiftsimio/initial_conditions/generate_particles.py", line 1111, in iteration_step
    h, rho = self.compute_h_and_rho()
  File "/Users/mphf18/Documents/GitHub/swiftsimio/swiftsimio/initial_conditions/generate_particles.py", line 1048, in compute_h_and_rho
    tree = KDTree(self.x, boxsize=boxsize)
  File "ckdtree.pyx", line 539, in scipy.spatial.ckdtree.cKDTree.__init__
ValueError: Some input data are greater than the size of the periodic box.
```

mladenivkovic:
Are you using the current version? It fails at line 555 for me.

mladenivkovic:
A small error snuck in in the past few commits. This should fix it, I was able to run a few steps with it.

swiftsimio/initial_conditions/generate_particles.py
mladenivkovic: Fine, have it your way.

JBorrow:
Excellent, thanks - that fixes it. At the moment it's quite slow, but I have a 10X speedup in my version that's now bounded by the tree search speed.

JBorrow:
We build the tree from `self.x` twice in `iteration_step` - once for computing h, rho, and the second time for use in the offset calculation - is there a specific reason for this?

mladenivkovic:
Yes. The tree doesn't track changes in `self.x`, so every time you update `self.x`, e.g. when redistributing a handful of particles, or moving all of them, you'll get bogus results if you don't rebuild the tree.

JBorrow:
Yes, but on line 1117 of `generate_particles.py` if `redistribute` is False, we've just built the tree twice for no reason.

mladenivkovic:
True. But I didn't think that would be a problem, typically the dumps are only every 100ish iterations. If you think that'll be a bottleneck, it's a simple fix with 
``` 
if dump_now and not redistribute:
    # build tree
    tree = KDTree(...)
```

JBorrow:
Right, but don't we only need to re-build the tree if there is a redistribute?

mladenivkovic:
Yes, but the dumping needs the smoothing lengths first. So if you dumped before, don't rebuild the tree unless we also redistributed in the same step.

JBorrow:
How frequent are `redistribute`s ?

mladenivkovic:
As frequent as you set them. I don't recommend going below every 20 iterations.

JBorrow:
Ok, but then we are double-tree-building when we don't need to almost every step, right?

JBorrow:
I only ask this because 10% of the run-time is currently spent in `compute_h_and_rho`

mladenivkovic:
What do you mean every step? Default values are redistribution every 20th step and dump every 50th step, is that almost every step for you?

JBorrow:
Oh :). I see how it works now. Odd that we spend so much time in that function, I will look at optimising it some more.

mladenivkovic:
Are we giving up on this?

JBorrow:
i want to come back to it, as it will be an excellent feature to have... but it needs to be significantly faster

PR #70, opened 2020-08-07
Minor formatting changes to IC

JBorrow:
@mladenivkovic can you check these to make sure they look ok to you? I've basically just moved a few things around to make some of the functions a little shorter.

### DIFF
--- a/swiftsimio/initial_conditions/generate_particles.py
+++ b/swiftsimio/initial_conditions/generate_particles.py
@@ -237,7 +237,9 @@ class IterData(object):
 
         return mips
 
-    def calculate_normalisation_constants(self,  min_delta_r_norm: float, delta_init: Optional[float] = None):
+    def calculate_normalisation_constants(
+        self, min_delta_r_norm: float, delta_init: Optional[float] = None
+    ):
         """
         Parameters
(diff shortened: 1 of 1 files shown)


swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -529,26 +675,13 @@ def initial_setup(
...
-            + Must be called AFTER the ``self.run_params`` paramters have been 
-              tweaked by the user.
+        + Must be called AFTER the ``self.run_params`` paramters have been 
+        tweaked by the user.
 
         """
-        # do this again in case some brainiac has the
-        # bright idea to change it in the meantime
-        self.npart = self.number_of_particles ** self.ndim
-        self.boxsize_to_use = self.boxsize.to(self.unit_system["length"]).value
```
mladenivkovic, line 544: I'm worried about this specifically because of the comment above.

JBorrow: Because this is now a @property it's calculated every time it's used

mladenivkovic: allrighty then

PR #73, opened 2020-08-25
vectorised a lot of operations, liberal use of numba kernels

JBorrow:
Hi @mladenivkovic - this includes significant speed improvements (over 10x) on your IC code. It still needs work but we're getting there!

### DIFF
--- a/swiftsimio/initial_conditions/generate_particles.py
+++ b/swiftsimio/initial_conditions/generate_particles.py
@@ -15,6 +15,14 @@ from copy import deepcopy
 from .IC_kernel import get_kernel_data
 from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
 from swiftsimio import Writer, load
+from swiftsimio.accelerated import jit
+
+@jit(nopython=True, fastmath=True)
+def correct_for_periodic(dx, ndim, boxsize):
+    for d in range(ndim):
+        bhalf = 0.5 * boxsize[d]
+        dx[dx[:, d] > bhalf, d] -= boxsize[d]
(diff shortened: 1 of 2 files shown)


swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -1206,34 +1239,54 @@ def iteration_step(
...
+            for p, (dist, neighs) in enumerate(zip(distances, neighbours)):
+                p += starting_index
+
+                # tree.query returns index npart where not enough neighbours are found
+                correct = neighs < self.npart
+                dist = dist[correct][1:]  # skip first neighbour: that's particle itself
+                neighs = neighs[correct][1:]
+
+                if neighs.shape[0] == 0:
+                    raise RuntimeError("Found no neighbour for a particle.")
```
mladenivkovic, line 1280: Again this might crash in 1276 when calling `dist = dist[correct][1:]`

swiftsimio/initial_conditions/generate_particles.py
```diff
@@ -15,6 +15,14 @@
 from .IC_kernel import get_kernel_data
 from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
 from swiftsimio import Writer, load
+from swiftsimio.accelerated import jit
+
+@jit(nopython=True, fastmath=True)
+def correct_for_periodic(dx, ndim, boxsize):
```
mladenivkovic, line 21: Some documentation would be great here that this is for distances, not particle positions

mladenivkovic (changes requested):

PR #77, opened 2020-10-05
added dust field

wmfw23:
added line to read the new-format dust fraction array for gas particles


JBorrow:
Hi James, this shouldn't be necessary any more - it should get picked up automatically.

I completely forgot about adding your dust table - I'll do that today once I've sorted out some admin stuff :)>

PR #83, opened 2020-11-10
Add scatter for gpu

loikki:
This is largely a copy/paste from the fast function.

### DIFF
--- a/swiftsimio/optional_packages.py
+++ b/swiftsimio/optional_packages.py
@@ -6,7 +6,10 @@ This includes:
 + tqdm: progress bars
 + scipy.spatial: KDTrees
 + sphviewer: visualisation
++ numba/cuda: visualisation
 """
+
+# TQDM
 try:
     from tqdm import tqdm
 
@@ -27,6 +30,7 @@ except (ImportError, ModuleNotFoundError):
     TREE_AVAILABLE = False
 
 
+# Py-sphviewer
 try:
     import sphviewer as viewer
 
@@ -36,6 +40,7 @@ except (ImportError, ModuleNotFoundError):
     SPHVIEWER_AVAILABLE = False
 
 
+# Astropy
 try:
     import astropy
 
@@ -45,11 +50,14 @@ except (ImportError, ModuleNotFoundError):
     ASTROPY_AVAILABLE = False
 
 
+# Numba/CUDA
 try:
     from numba.cuda.cudadrv.error import CudaSupportError
 
     try:
         import numba.cuda.cudadrv.driver as drv
+        from numba import cuda
+        from numba.cuda import jit as cuda_jit
 
         d = drv.Driver()
         d.initialize()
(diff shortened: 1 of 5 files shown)


swiftsimio/visualisation/projection_backends/__init__.py
```diff
@@ -59,3 +59,17 @@
...
+        scatter_parallel as scatter_gpu_parallel,
+    )
+
+    backends["gpu"] = scatter_gpu
+    backends_parallel["gpu"] = scatter_gpu_parallel
+except ImportError:
+    print(
+        "Unable to load the GPU module. Please check the module numba.cuda "
+        "if you wish to use them."
+    )
```
JBorrow, line 75: Please do this import trickery in optional_packages: https://github.com/SWIFTSIM/swiftsimio/blob/master/swiftsimio/optional_packages.py

JBorrow (commented):
Apart from import stuff this looks good to me. Backends should always contain `gpu`.

swiftsimio/optional_packages.py
```diff
@@ -7,6 +7,7 @@
 + scipy.spatial: KDTrees
 + sphviewer: visualisation
 """
+from numba.cuda.cudadrv.error import CudaSupportError
```
JBorrow, line 10: Hmm this is quite tricky; the user may not have numba installed at all (hence it itself is an optional package). Can we mock this up?

swiftsimio/optional_packages.py
```diff
@@ -43,4 +44,17 @@ def tqdm(x, *args, **kwargs):
...
+    import numba.cuda.cudadrv.driver as drv
+
+    d = drv.Driver()
+    d.initialize()
+
+    CUDA_AVAILABLE = True
+except CudaSupportError:
+    CUDA_AVAILABLE = False
+
+print("CUDA found?", CUDA_AVAILABLE)
```
JBorrow, line 60: Print statements? Smh :p

swiftsimio/visualisation/projection_backends/gpu.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Notes
+    -----
+
+    Explicitly defining the types in this function allows
+    a performance improvement.
+    """
+    if not CUDA_AVAILABLE:
+        raise Exception(
+            "Unable to load the GPU function. " "Please check your module numba.cuda."
```
JBorrow, line 217: Can we raise the CudaSupportError here?

swiftsimio/optional_packages.py
loikki: Sorry, I forgot to commit this file

swiftsimio/visualisation/projection_backends/gpu.py
loikki: For the tests it will become a problem. We need to know in advance if the function is available or not.

swiftsimio/optional_packages.py
loikki: When I check for cuda, I am catching an additional exception now.

JBorrow:
Happy for me to take a look in the branch to resolve some of the optional module madness?

loikki:
Yes feel free to modify it

JBorrow:
Can you check that you're happy with the changes I made in the last four commits?

Sorry for being annoying about this, I just need to be absolutely sure that this isn't going to break on people's computers who have an old version of numba installed and things.

loikki:
I am happy with the changes, but you can reduce the amount of copy/paste by taking out of the `try: except:` the definition of `cuda_jit` and define it if cuda is not available.

JBorrow:
Thanks so much for this Loic! It would be great to see some images that you've produced with it :).

Issue #115, opened 2022-03-14
Non-cubic boxes not supported for slice visualisation


zorrybelcheva:
Hi guys,

To my understanding it's currently not possible to make non-square slice images directly. The error returned when e.g. slice_gas() is prompted with a non-cubic box is `Slice code is currently not able to handle non-square images`. 

A way to bypass this is to use slice_scatter(), rescaling the coordinates to [0, 1] and then cutting off the empty parts of the image, or alternatively to stitch together a grid of images, e.g. 2 along the x-direction if using a box of dimensions [2, 1, 1]. I'm wondering if this would be the standard way to approach this?

bwvdnbro:
`slice_gas()` calls `slice_scatter()` internally, so using that function is definitely appropriate. I think the first option (where you rescale and then cut the empty parts of the image) should be faster, so I would recommend that.

I am wondering why this restriction exists. There is no fundamental reason why non-square images would not work. The only problem is that you then need two numbers for the number of pixels, and you risk having incompatible dimensions in the actual image size and the image pixel size.

A quick workaround would be to always produce a square image that has the maximum of the two dimensions as side length. We could then potentially automatically cut empty pixels. That would be the equivalent of your first option.

@JBorrow any thoughts?

bwvdnbro:
@zorrybelcheva you can find a tentative implementation that supports non-cubic boxes in the branch `non-square-slices`. We are still figuring out some subtleties when you use rotations, but if you don't need rotations you can already give this a try.

JBorrow:
Is this now fixed in v5.0.0?

MatthieuSchaller:
@zorrybelcheva could you confirm whether this now works in the latest version?

zorrybelcheva:
Yes! Can confirm there is no error anymore.

PR #116, opened 2022-03-16
Added support for non-square slices.

bwvdnbro:
As discussed in #115, support for non-square images was lacking from `visualisation/slice.py`.

I have provided a minimal form of support that avoids some of the potential pitfalls: instead of normalising the pixel coordinates using the range in both the x and y dimension, the maximum of these ranges is used. A square image with the requested resolution is then produced that will contain empty pixels in the smaller dimension. These empty pixels are filtered out before returning the non-square result image.

This minimal implementation implies that
 - the image resolution argument is applied to the longer dimension only (after possible rotations) and should be interpreted as such
 - there could be a small mismatch in the pixel length of the smaller dimension because of rounding down
 - no stretching happens, because all coordinates are normalised with the same number

### DIFF
--- a/swiftsimio/visualisation/slice.py
+++ b/swiftsimio/visualisation/slice.py
@@ -358,7 +358,7 @@ def slice_gas_pixel_grid(
     x_range = x_max - x_min
     y_range = y_max - y_min
 
-    # Deal with non-square boxes:
+    # Deal with non-cubic boxes:
     # we always use the maximum of x_range and y_range to normalise the coordinates
     # empty pixels in the resulting square image are trimmed afterwards
     max_range = max(x_range, y_range)
@@ -383,10 +383,10 @@ def slice_gas_pixel_grid(
     common_parameters = dict(
         x=(x - x_min) / max_range,
         y=(y - y_min) / max_range,
-        z=z / box_z,
+        z=z / max_range,
         m=m,
         h=hsml / max_range,
-        z_slice=slice,
+        z_slice=slice * box_z / max_range,
         res=resolution,
     )
 
@@ -488,18 +488,16 @@ def slice_gas(
     if region is not None:
         x_range = region[1] - region[0]
         y_range = region[3] - region[2]
-        units = 1.0 / (x_range * y_range * data.metadata.boxsize[2])
+        max_range = max(x_range, y_range)
+        units = 1.0 / (max_range ** 3)
         # Unfortunately this is required to prevent us from {over,under}flowing
         # the units...
         units.convert_to_units(
             1.0 / (x_range.units * y_range.units * data.metadata.boxsize.units)
         )
     else:
-        units = 1.0 / (
-            data.metadata.boxsize[0]
-            * data.metadata.boxsize[1]
-            * data.metadata.boxsize[2]
-        )
+        max_range = max(data.metadata.boxsize[0], data.metadata.boxsize[1])
+        units = 1.0 / (max_range ** 3)
         # Unfortunately this is required to prevent us from {over,under}flowing
         # the units...
         units.convert_to_units(1.0 / data.metadata.boxsize.units ** 3)
--- a/swiftsimio/visualisation/slice.py
+++ b/swiftsimio/visualisation/slice.py
@@ -17,7 +17,8 @@ from numpy import (
     matmul,
     copy,
 )
-from unyt import unyt_array
+from unyt import unyt_array, unyt_quantity
+import unyt
 from swiftsimio import SWIFTDataset
 
 from swiftsimio.accelerated import jit, prange, NUM_THREADS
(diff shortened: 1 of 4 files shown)


JBorrow:
Please add a relevant test and can you show an example image (e.g. the blob test?)

bwvdnbro:
I have added a test.

This is what it looks like when applied to the output of the 3D Sod shock:
gas_slice_map

Script:
```
from swiftsimio import load
from swiftsimio.visualisation.slice import slice_gas
from swiftsimio.visualisation.rotation import rotation_matrix_from_vector

data = load("sodShock_0001.hdf5")

# This creates a grid that has units msun / Mpc^3, and can be transformed like
# any other unyt quantity. Note that `slice` is given in terms of the box-size,
# so here we are taking a slice at z = boxsize / 2.
mass_map = slice_gas(
    data,
    slice=0.5,
    resolution=1024,
    project="masses",
    parallel=True,
)

# Let's say we wish to save it as g / cm^2,
from unyt import g, cm

mass_map.convert_to_units(g / cm**3)

from matplotlib.pyplot import imsave
from matplotlib.colors import LogNorm

# Normalize and save
imsave("gas_slice_map.png", LogNorm()(mass_map.value), cmap="viridis")
```

JBorrow:
Final question is are you sure this doesn't affect the units in slice_gas?

bwvdnbro:
That question opened a whole can of worms...

Short answer: no.

Long answer: the units are wrong and I don't think the slicing function works correctly if `x_range` and `y_range` are not equal to the box size in `z`, even in the old version. The reason is that `slice_scatter` assumes that the re-scaling from coordinates to `[0,1]` is shape-preserving (because the re-scaled kernel should still be spherical), which really only works if the denominator in the re-scaling is the same for all 3 coordinate directions. The old version only guaranteed `x_range==y_range` and completely ignored `z`. That could explain why the Sod shock image looks so noisy (while the Sod curves are reasonably well behaved).

This would be pretty easy to fix if there were no rotations. With rotations, the length in `z` is ill-defined and I don't really know how to fix this.

bwvdnbro:
I just pushed a new version that fixes the units, but will not work properly with rotations. @JBorrow I would really like your input on this to decide what to do about the `z` direction.

Updated Sod shock image:
gas_slice_map

JBorrow:
Well to me it seems like `z` should have no bearing on the output _units_, as they're 3D? But I certainly see the issue with rotations.

I think we could make this a physical offset from the centre (which is x,y,z = 0.5 * boxsize by default)?

bwvdnbro:
I agree that it's not really the _units_ per se. However, because of the re-scaling, the map is always produced in units of [Q]/(Lx Ly Lz), where Lx, Ly, Lz are the box size or region size in the different directions and [Q] is the unit of the mapped quantity. What I'm arguing here is that Lx, Ly and Lz should always be the same to conserve shapes when you do the re-scaling (e.g. the shape of the _spherical_ smoothing kernel). Right now, Lz is always chosen to just be the box size in z, without any restrictions on its value. While it really should be Lz=L, with L = max(Lx, Ly). I have changed that in the last commit.

JBorrow:
I'm not sure that's the case. Here we only use `z` and `h` to compute the distance from the plane... So as long as the `z` units and `h` units are the same, you're good to go, no?

bwvdnbro:
Yes. So the `z` and `h` units have to be the same to check the overlap with the slice, and the `x`, `y` and `h` units also have to be the same for the kernel sampling to work. Logic then dictates that `x`, `y` and `z` have to have the same units, so even though you don't care about `z`, you still need to make sure the re-sampling of `x` and `y` conserves 3D shapes.

The only difference between the slice re-sampling and the projection re-sampling is that you don't care about the re-sampling factor for `z`. So the box size in `z` does not enter the equations. Is that what you mean?

bwvdnbro:
Since @Fonotec has no found an example that returns the wrong density in a slice (because of the wrong scale factor for the z coordinates), I had another go at cleaning this up. I have now changed the meaning (and name) of the `slice` argument to the functions so that it is a physical offset rather than some fraction of the box size. I thought a sensible default would be the origin of the box, since that seems more natural than the center if you use particle coordinates to pick the location of the slice. If a rotation center is provided, that is used instead, since you probably want to put your slice close to the center of an object you just rotated. We can argue about those choices.

I had to update the unit tests to reflect these changes, and I suspect the documentation also needs to be updated.

I think we might need to release a new version once these changes are made, since the slicing functionality is effectively broken if you do not use the full (cubic) box.

swiftsimio/visualisation/slice.py
```diff
@@ -396,13 +401,18 @@ def slice_gas_pixel_grid(
...
+
+    # trim the image to remove empty pixels
+    return image[:xres, :yres]
 
 
 def slice_gas(
     data: SWIFTDataset,
     resolution: int,
-    slice: float,
+    z_slice: unyt_quantity = 0.0 * unyt.m,
```
JBorrow, line 415: I don't like this - can we have `z_slice: Optional[unyt_quantity] = None` and set it to `0.0 * data.gas.coordinates.units`?

JBorrow:
I'm happy to merge this after there have been associated documentation improvements.

bwvdnbro:
I think this should be ready to go now.

PR #118, opened 2022-03-31
Writer: added check whether coordinates and velocities arrays have same shape

mladenivkovic:
This adds a check whether the `coordinates` and `velocities` arrays of each particle have the same shape.

Currently the `coordinates` and `velocities` arrays are part of the `shared` portion of the required fields for all particles, so this should work for any particle type.

I ran into this issue when generating ICs, where the velocity fields would get filled up with random junk at times while running swift with those ICs, leading to crashes and errors that looked like race conditions / dependency issues.


bwvdnbro:
Seems like a sensible thing to check. Thanks!

Issue #119, opened 2022-04-06
Pickling cosmo_array objects strips attributes


jchelly:
I've just found that pickling and then un-pickling a cosmo_array strips the cosmo_factor and comoving attributes. This is a problem when using mpi4py's ability to send and receive data structures by pickling them because some of the metadata is lost.

For example:
```
import pickle
import unyt
import swiftsimio.objects as o

a = 1.0
a_scale_exponent = 1.0
cosmo_factor = o.cosmo_factor(o.a**a_scale_exponent, a)
arr1 = o.cosmo_array(unyt.unyt_array(1.0, units=unyt.cm), cosmo_factor=cosmo_factor, comoving=True)
data = pickle.dumps(arr1)
arr2 = pickle.loads(data)
``` 
Here, arr1.cosmo_factor and arr1.comoving are defined but the same attributes are missing from arr2.

jchelly:
I think the fix for this might be to define __getstate__ and __setstate__ for cosmo_arrays. Unyt already does this - it calls the get/set state for the underlying numpy array and then deals with it's own extra bits of state.

jchelly:
Fixed by #120.

PR #120, opened 2022-04-06
Allow pickling of cosmo_arrays

jchelly:
Currently pickling cosmo_arrays strips their cosmo_factor and comoving attributes.  This pull request adds `__reduce__` and `__setstate__` methods which preserve these attributes.


jchelly:
This should fix #119.

JBorrow:
Thanks, this is excellent! 

Note that the `name` attribute is not dumped by `unyt`, but that's an upstream problem.

Issue #74, opened 2020-08-27
Import of smoothing_length_generation does not work as expected


Fonotec:
importing visualisation as:
`import swiftsimio.visualisation as swvis`
and than generating the smoothing lengths as:
`swvis.smoothing_length_generation.generate_smoothing_lengths()`
does not find the function while:
`from swiftsimio.visualisation.smoothing_length_generation import generate_smoothing_lengths`
does work.

Fonotec:
I don't know, after reporting this I never imported the functions different than the last line here.

PR #126, opened 2022-04-20
Import smoothing_length_generation is as part of swiftsimio.visualisation

bwvdnbro:
Addresses #74.


JBorrow:
This doesn't make much sense to me

JBorrow:
If we're going to do this, we might as well just

`from .smoothing_length_generation import generate_smoothing_lengths`, and make it a top-level function.

bwvdnbro:
Works for me too. I agree with @Fonotec that you should be able to access all the functionality with one import.

bwvdnbro:
I have implemented your suggestion.

PR #132, opened 2022-05-27
Cosmo array ufuncs

### LINKED ISSUE #131: cosmo_array resets extra attributes to defaults when negated
```
>>> import unyt as u
>>> from swiftsimio.objects import cosmo_array, cosmo_factor, a
>>> (-cosmo_array(np.ones(5), u.Mpc, comoving=False, cosmo_factor=cosmo_factor(a**1, 1.0))).comoving  # comoving set False, so expect False
True
```
The `cosmo_factor` attribute is also reset to `None`. I guess this probably applies to (many?) more operations than negation, haven't investigated further though.

### LINKED ISSUE #142: cosmo_array multiplication does not return cosmo_array
Merging #137 and #141 unfortunately broke some of the visualisation functions. Consider the following code snippet:
```
data.gas.HI_mass_fraction = (
    data.gas.masses
    * data.gas.element_mass_fractions.hydrogen
    * data.gas.species_fractions.HI
)
HI_map = project_gas(
                    data,
                    resolution=8000,
                    project="HI_mass_fraction",
                )
```
All the `data.gas.XXX` attributes are `cosmo_array`s, but unfortunately `data.gas.HI_mass_fraction` is not, since multiplication of `cosmo_array`s does not conserve the `cosmo` properties. `project_gas` will fail when it tries to confirm that `HI_mass_fraction` and the coordinates are compatible cosmological quantities.

The only way to fix this would be to add support for multiplication (and maybe some other operators) to `cosmo_array`. We probably also want to disable unsupported operations to be on the safe side.

kyleaoman:
Implementation of support for numpy ufuncs so that results preserve sensible `comoving` and `cosmo_factor` attributes. Will resolve #131 and resolve #142, which are symptoms of the previous lack of support for this.

### DIFF
--- a/swiftsimio/objects.py
+++ b/swiftsimio/objects.py
@@ -140,9 +140,11 @@ def _multiply_cosmo_factor(ca_cf1, ca_cf2, **kwargs):
         # two is not a cosmo_array, allow e.g. multiplication by constants:
         return cf1
     elif (ca1 and ca2) and ((cf1 is None) or (cf2 is None)):
-        # both cosmo_array but not both with cosmo_factor (both without shortcircuited above already):
+        # both cosmo_array but not both with cosmo_factor
+        # (both without shortcircuited above already):
         warnings.warn(
-            f"Mixing ufunc arguments with and without cosmo_factors ({cf1} and {cf2}), discarding cosmo_factor in return value.",
+            f"Mixing ufunc arguments with and without cosmo_factors ({cf1} and {cf2}),"
+            f" discarding cosmo_factor in return value.",
             RuntimeWarning,
         )
         return None
@@ -169,16 +171,20 @@ def _preserve_cosmo_factor(ca_cf1, ca_cf2=None, **kwargs):
         # only one is cosmo_array
         return cf2
     elif (ca1 and ca2) and (cf1 is None and cf2 is not None):
-        # both cosmo_array, but not both with cosmo_factor (both without shortcircuited above already):
+        # both cosmo_array, but not both with cosmo_factor
+        # (both without shortcircuited above already):
         warnings.warn(
-            f"Mixing ufunc arguments with and without cosmo_factors, continuing assuming provided cosmo_factor ({cf2}) for all arguments.",
+            f"Mixing ufunc arguments with and without cosmo_factors, continuing assuming"
+            f" provided cosmo_factor ({cf2}) for all arguments.",
             RuntimeWarning,
         )
         return cf2
     elif (ca1 and ca2) and (cf1 is not None and cf2 is None):
-        # both cosmo_array, but not both with cosmo_factor (both without shortcircuited above already):
+        # both cosmo_array, but not both with cosmo_factor
+        # (both without shortcircuited above already):
         warnings.warn(
-            f"Mixing ufunc arguments with and without cosmo_factors, continuing assuming provided cosmo_factor ({cf1}) for all arguments.",
+            f"Mixing ufunc arguments with and without cosmo_factors, continuing assuming"
+            f" provided cosmo_factor ({cf1}) for all arguments.",
             RuntimeWarning,
         )
         return cf1
@@ -238,13 +244,12 @@ def _passthrough_cosmo_factor(ca_cf, ca_cf2=None, **kwargs):
             f"Ufunc arguments have cosmo_factors that differ: {cf} and {cf2}."
         )
     else:
-        # passthrough is for e.g. ufuncs with a second dimensionless argument, so ok if cf2 is None and cf1 is not
+        # passthrough is for e.g. ufuncs with a second dimensionless argument,
+        # so ok if cf2 is None and cf1 is not
         return cf
 
 
-def _return_without_cosmo_factor(
-    ca_cf, ca_cf2=None, inputs=None, zero_comparison=False
-):
+def _return_without_cosmo_factor(ca_cf, ca_cf2=None, inputs=None, zero_comparison=None):
     ca, cf = ca_cf
     ca2, cf2 = ca_cf2 if ca_cf2 is not None else (None, None)
     if ca_cf2 is None:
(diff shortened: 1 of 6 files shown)


kyleaoman:
The extra attributes of `cosmo_array`s should now propagate correctly through all numpy ufuncs that are supported by `unyt`. I've tried to be diligent about covering these with unit tests without being pedantic about testing every single ufunc. Also made some tweaks to comparison operators between `cosmo_factors` to try and make them more reliable (though there are still some corner cases that aren't covered, I don't know how to use `sympy` to symbolically compare the expressions). Guess there might be some questions about the changes here, happy to talk it over...

bwvdnbro:
I get
```
would reformat /net/winkel/data1/vandenbroucke/swiftsimio/swiftsimio/objects.py
All done! 💥 💔 💥
1 file would be reformatted, 67 files would be left unchanged.
ERROR: needs formatting
```

Could you share the output of `black_formatting_env/bin/pip3 list`? We enforce a specific `black` version exactly to avoid this kind of problem, so this is really strange.

kyleaoman:
@bwvdnbro I did check the black version and as far as I can tell it matches...

```
koman@farseer> black_formatting_env/bin/pip3 list
Package    Version
---------- -------
appdirs    1.4.4
attrs      21.4.0
black      19.3b0
click      8.0.4
pip        21.2.3
setuptools 57.4.0
toml       0.10.2
WARNING: You are using pip version 21.2.3; however, version 22.1.2 is available.
You should consider upgrading via the '/home/koman/code/swiftsimio/black_formatting_env/bin/python3 -m pip install --upgrade pip' command.
```

bwvdnbro:
All those versions match. What about `python3 --version`?

This is what my version wants to change:
```
diff --git a/swiftsimio/objects.py b/swiftsimio/objects.py
index 06d7dfc..016f9ec 100644
--- a/swiftsimio/objects.py
+++ b/swiftsimio/objects.py
@@ -116,7 +116,9 @@ def _propagate_cosmo_array_attributes(func):
 
 
 def _sqrt_cosmo_factor(ca_cf, **kwargs):
-    return _power_cosmo_factor(ca_cf, (False, None), power=.5)  # ufunc sqrt not supported
+    return _power_cosmo_factor(
+        ca_cf, (False, None), power=0.5
+    )  # ufunc sqrt not supported
 
 
 def _multiply_cosmo_factor(ca_cf1, ca_cf2, **kwargs):
@@ -195,7 +197,7 @@ def _power_cosmo_factor(ca_cf1, ca_cf2, inputs=None, power=None):
             raise ValueError("Exponent must be dimensionless.")
         else:
             power.to_value(unyt.dimensionless)
-    if ca2 and cf2.a_factor != 1.:
+    if ca2 and cf2.a_factor != 1.0:
         raise ValueError("Exponent has scaling with scale factor != 1.")
     if cf1 is None:
         return None
@@ -942,9 +944,7 @@ class cosmo_array(unyt_array):
                 )
             else:
                 ret_cf = _power_cosmo_factor(
-                    cfs[0],
-                    (False, None),
-                    power=power_sign * inputs[0].size,
+                    cfs[0], (False, None), power=power_sign * inputs[0].size
                 )
         else:
             ret_cf = self._cosmo_factor_ufunc_registry[ufunc](*cfs, inputs=inputs)
```
Seems pretty random.

kyleaoman:
I'm on python3.10 (including in the formatting venv). Two of those changes look like they might be triggered by a change to the default line length. The other one `1. -> 1.0` I'm a bit surprised by.

It gets stranger, though. I set up a new venv with python3.8.2 - as far as I can tell this is the version github is using, as it's default on Ubuntu 20.04 and I don't see a python version specified anywhere in the formatting workflow. And it's still reporting
```
Formatting all Python scripts
All done! ✨ 🍰 ✨
68 files left unchanged.
```

kyleaoman:
Rebased on the current upstream master branch and cleaned up some extraneous merges in the commit history. This is all set to go now from my end, and tests pass.

A couple of notes:
 - There's a bug in unyt that affects the (very obscure) use of the reduce method of the np.divide ufunc. I fixed that in unyt PR#231, but it hasn't gone to a release yet, and the last unyt release is actually from 2020 at this point. I've included a test marked xfail that will pass (and thus count as a failure) when the fix for unyt comes through, and left some comments signposting what needs to be done when that happens.
 - I've erred on the side of aggressively emitting warnings when a ufunc with >1 argument is used on a combination of a cosmo_array and a unyt_array (or ndarray in some cases). I still let the calculation proceed assuming the cosmo_array attributes of the argument that has them. This has some potentially undesirable side effects, notably that multiplication by a constant from the unyt.physical_constants module complains loudly, even though this should be allowed. Unfortunately there's nothing distinguishing those constants from any other unyt_quantity, so the case can't be handled separately for now.
 - We currently don't have a cosmo_quantity analogue of a unyt_quantity. There are places where I wished these existed, but I wasn't ready to open that can of worms for this PR, so I've just worked around that lack for now.

swiftsimio/objects.py
```diff
@@ -26,6 +115,199 @@ def wrapped(self, *args, **kwargs):
...
+        )
+        return cf1
+    elif (ca1 and ca2) and (cf1 != cf2):
+        raise ValueError(
+            f"Ufunc arguments have cosmo_factors that differ: {cf1} and {cf2}."
+        )
+    elif (ca1 and ca2) and (cf1 == cf2):
+        return cf1  # or cf2, they're equal
+    else:
+        raise NotImplementedError  # should not get here
```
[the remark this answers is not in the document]
kyleaoman: Indeed, fixed.

swiftsimio/objects.py
```diff
@@ -26,6 +121,243 @@ def wrapped(self, *args, **kwargs):
...
+    if inputs is not None and power is not None:
+        raise ValueError
+    ca1, cf1 = ca_cf1
+    ca2, cf2 = ca_cf2
+    power = inputs[1] if inputs else power
+    if hasattr(power, "units"):
+        if not power.units.is_dimensionless:
+            raise ValueError("Exponent must be dimensionless.")
+        else:
+            power.to_value(unyt.dimensionless)
```
JBorrow, line 211: `to_value` returns a copy of the array converted to that unit. It does not affect the array itself... Perhaps `convert_to_value` is what you were looking for? Though this does mutate the external array's state which may be bad behavior. Can you just leave it in its original (dimensionless compatible) units?

kyleaoman: I think we're stuck with a copy operation here. This is determining what exponent to raise the units to. `convert_to_value` is a bad idea because it can unexpectedly modify the user's input data. Taking the raw value risks something like `power = 2000 * u.mm / (1 * u.m)`, when that gets used to exponentiate the units you'll get an exponent of `2000` instead of `2`. That leaves copying. I don't think it's a big deal because the thing being copied is the exponent argument of `np.power`, which in most cases is a scalar, and already dimensionless. Only if you had something like `np.power(2, np.arange(100000) * u.m / u.cm)` would you get an expensive copy, and it's hard to think of a use-case for such an expression. I guess I could add an `elif power.units is u.dimensionless` to avoid the copy if the units are explicitly the dimensionless unit?

JBorrow: Well the problem here is that in the `else` case you're not actually using the converted value... You can do:

`use_power = power.to_value(unyt.dimensionless)`

but as the conversion is not done in-place right now this line does nothing. I think adding the elif for the simple case is a good idea.

JBorrow: Well the problem here is that in the `else` case you're not actually using the converted value... You can do:

`use_power = power.to_value(unyt.dimensionless)`

but as the conversion is not done in-place right now this line does nothing. I think adding the elif for the simple case is a good idea.

kyleaoman: Um, yes, you're right of course. Sorry I missed the point on that one before, fixed now.

PR #133, opened 2022-06-01
make particle IDs start at 1

mladenivkovic:
So apparently when running with DM and hydro, particle ID = 0 is a no-no. So let's just always start with 1.


bwvdnbro:
Could you add a comment to the script explaining why we start at 1? Just to avoid accidentally reverting this in the future.

mladenivkovic:
Which script are you referring to?

mladenivkovic:
Gotcha. Done.

PR #136, opened 2022-06-10
Zero size array support

bwvdnbro:
https://gitlab.cosma.dur.ac.uk/swift/swiftsim/-/merge_requests/1560 adds support for empty particle arrays to SWIFT. These arrays can be present if a simulation is run with a certain particle type enabled (e.g. stars), but a snapshot is written before any of those particles were actually created. The advantage of writing empty arrays is that then all snapshots for that run will contain exactly the same datasets and metadata.

This PR adds support for this to `swiftsimio`. We read the new header attribute `CanHaveTypes` and use this rather than the particle number to decide whether or not to create the particle groups. If this attribute is not present, we revert to the old behaviour.

Scripts that are run on every snapshot of a run should now always get the same `SWIFTDataset`, regardless of whether or not a certain particle type is actually present in that snapshot.

### DIFF
--- a/swiftsimio/reader.py
+++ b/swiftsimio/reader.py
@@ -599,7 +599,10 @@ class SWIFTMetadata(object):
         The particle types that are present in the file.
         """
 
-        return np.where(np.array(self.num_part) != 0)[0]
+        if hasattr(self, "has_type"):
+            return np.where(np.array(self.has_type) != 0)[0]
+        else:
+            return np.where(np.array(self.num_part) != 0)[0]
 
     @property
     def present_particle_names(self):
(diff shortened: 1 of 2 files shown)


MatthieuSchaller:
Seems straightforward enough. I guess this means we expose a zero-sized array for all the properties of the yet-unborn stars for instance.

swiftsimio/reader.py
```diff
@@ -599,7 +599,10 @@ def present_particle_types(self):
         The particle types that are present in the file.
         """
 
-        return np.where(np.array(self.num_part) != 0)[0]
+        if hasattr(self, "has_type"):
+            return np.where(np.array(self.has_type) != 0)[0]
+        else:
+            return np.where(np.array(self.num_part) != 0)[0]
```
JBorrow, line 605:
```suggestion
        return np.where(np.array(getattr(self, "has_type", self.num_part)) != 0)[0]
```

bwvdnbro:
I have merged in the new formatting script and applied it. Should be ready to go now.

MatthieuSchaller (approved):

Issue #135, opened 2022-06-10
Astropy deprecation warning in conversions.py


bwvdnbro:
I am getting the following warnings when using `swiftsimio` with `astropy 5.1`:
```
WARNING: AstropyDeprecationWarning: `astropy.cosmology.flrw.a_B_c2` is a private variable (since v5.1) and in future will raise an exception. [astropy.cosmology.flrw]
WARNING: AstropyDeprecationWarning: `astropy.cosmology.flrw.critdens_const` is a private variable (since v5.1) and in future will raise an exception. [astropy.cosmology.flrw]
```
Unfortunately, I couldn't find another way to obtain these variables through the API. I'll just signal it here, so that someone more knowledgeable about `conversions.py` can look at this.

bwvdnbro:
Why was this closed? #139 has not been merged and I still get this warning when I run the latest version of swiftsimio.

MatthieuSchaller:
Sorry, I misread the message above that said "fixed #135".

PR #139, opened 2022-06-14
Do not use soon to be private astropy attributes in conversions.

bwvdnbro:
As mentioned in #135, astropy will soon no longer allow access to some of the cosmology attributes that are used in `conversions.py` to convert from Omega_r to a CMB temperature.

To prevent any problems in the future, I have simply extracted the relevant expressions from astropy and added them directly to the script.


bwvdnbro:
Some more thoughts on this: it looks like the SWIFT header actually stores a CMB temperature these days. Can't we just use that instead of computing it? Or is that somehow not compatible?

JBorrow:
We could, yes. I'm not sure it was there when I initially wrote this code.

PR #144, opened 2022-07-07
Return cosmo_array from smoothing length calculation for non-gas particles

### LINKED ISSUE #143: Smoothing length calculation for non-gas particles returns `unyt_array`
Some recent changes to the visualisation module check that the smoothing lengths and coordinates have compatible attributes (e.g. both comoving, or both physical). However, `visualisation.smoothing_length_generation.generate_smoothing_lengths` returns a `unyt_array`, so these checks always fail with an `AttributeError` and the smoothing lengths are unusable.

Putting together a PR for this, it includes a test that currently fails:

```
@requires("cosmological_volume.hdf5")
def test_nongas_smoothing_lengths(filename):
    """
    Test that the visualisation tools to calculate smoothing lengths give usable results.
    """

    data = load(filename)
    data.dark_matter.smoothing_length = generate_smoothing_lengths(
        data.dark_matter.coordinates, data.metadata.boxsize, kernel_gamma=1.8
    )
    project_pixel_grid(
        data.dark_matter,
        boxsize=data.metadata.boxsize,
        resolution=256,
        project="masses",
    )

    # if project_pixel_grid runs without error the smoothing lengths seem usable
    return
```

kyleaoman:
Resolves #143

### DIFF
--- a/swiftsimio/visualisation/smoothing_length_generation.py
+++ b/swiftsimio/visualisation/smoothing_length_generation.py
@@ -12,7 +12,7 @@ from typing import Union
 
 
 def generate_smoothing_lengths(
-    coordinates: cosmo_array,
+    coordinates: Union[unyt_array, cosmo_array],
     boxsize: Union[unyt_array, cosmo_array],
     kernel_gamma: float32,
     neighbours=32,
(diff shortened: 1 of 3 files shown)


JBorrow:
This function is used in a lot of external code so I'm concerned about introducing breaking changes (in particular requiring cosmo array). I currently use this with Arepo data and a base unyt array in some scripts. Can we give the cosmo array sensible defaults?

kyleaoman:
Would we be happy returning a cosmo_array for cosmo_array input, and a unyt_array for unyt_array input? That should be simple to implement, and give intuitive results?

JBorrow:
Yeah that makes sense to do.

kyleaoman:
Great, all done.

PR #145, opened 2022-07-14
Adjustments for changes in unyt.

kyleaoman:
unyt has released a new version with a bugfix affecting cosmo_array. I've made the functional changes to handle the changes in unyt, and for now put a try/except on an import prompting users to update unyt if needed. If we prefer something more backwards compatible, that could be done, just let me know.

It also appears at a glance that python3.6 fails to install the latest unyt release during the test suite. Not sure what's going on with that, will have to look into it.

### DIFF
--- a/swiftsimio/objects.py
+++ b/swiftsimio/objects.py
@@ -8,6 +8,7 @@ import warnings
 import unyt
 from unyt import unyt_array
 from unyt.array import multiple_output_operators
+
 try:
     from unyt.array import POWER_MAPPING
 except ImportError:
--- a/swiftsimio/objects.py
+++ b/swiftsimio/objects.py
@@ -7,13 +7,11 @@ import warnings
 
 import unyt
 from unyt import unyt_array
-
-# Eventually import of POWER_SIGN_MAPPING will fail because it has been
-# replaced with POWER_MAPPING in a bugfix (unyt PR#231). Then see segment
-# of unyt.array.__array_ufunc__ where it is used and update
-# cosmo_array.__array_ufunc__ accordingly. There is an xfailing test in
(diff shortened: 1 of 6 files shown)


kyleaoman:
Failing tests is due to #146. This is otherwise ready to go (pending some input on how backwards compatible it should be). I could update setup.py for unyt>=2.9.0 and python>=3.8 here if we want to go that route.

swiftsimio/objects.py
```diff
@@ -7,13 +7,12 @@
...
-# Eventually import of POWER_SIGN_MAPPING will fail because it has been
-# replaced with POWER_MAPPING in a bugfix (unyt PR#231). Then see segment
-# of unyt.array.__array_ufunc__ where it is used and update
-# cosmo_array.__array_ufunc__ accordingly. There is an xfailing test in
-# test_cosmo_array_attrs to catch when this occurs with a test failure.
-from unyt.array import multiple_output_operators, POWER_SIGN_MAPPING
+try:
+    from unyt.array import POWER_MAPPING
+except ImportError:
+    raise ImportError("unyt >=2.9.0 required")
```
JBorrow, line 15: Change this in the setup.py too.

kyleaoman: Done, removed testing on python3.6 and updated mentions of the python version in the docs, too.

kyleaoman:
I think this is ready to go now. I think anyone who tries to use python<=3.6 will get an error from swiftsimio because it will look for things in unyt that don't exist. That should be clear that they need to get unyt 2.9.0, which in turn won't be available on older python versions, so I think it's adequately signposted. The swiftsimio docs also now updated to mention that python>=3.8 is required.

PR #147, opened 2022-07-26
Added support for non-square projections.

bwvdnbro:
Should be the equivalent of #116, but then for the projection function.

### DIFF
--- a/swiftsimio/visualisation/projection.py
+++ b/swiftsimio/visualisation/projection.py
@@ -184,11 +184,10 @@ def project_pixel_grid(
     x_range = x_max - x_min
     y_range = y_max - y_min
 
-    # Test that we've got a square box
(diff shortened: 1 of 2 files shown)


bwvdnbro:
The CI failures seem to be unrelated to this pull request.

JBorrow:
@kyleaoman failures seem to be due to new imports from unyt, can you check that out?

JBorrow:
Your changes look good to me @bwvdnbro

kyleaoman:
> @kyleaoman failures seem to be due to new imports from unyt, can you check that out?

Yup, there's a PR for this #145. No technical hurdles on that one, but need a bit of input on backwards compatibility and possibly to drop support for python3.6 (which unyt no longer supports) before merging.

PR #148, opened 2022-07-29
Made sure visualisation functions include periodic copies of particles.

bwvdnbro:
The visualisation functions only consider one periodic copy of each particle. As a result, there are some weird edge effects when visualising the full box. This is what you get when you visualise the full `cosmological_volume.hdf5` test file ("zoom" is the region inside the white square that excludes the edge):

Projection:
project

Slice:
slice

Volume render (slice through the centre of the voxel cube):
render

This PR adds periodic copies of all particles that could potentially contribute before calling the visualisation function. This restores the correct edge:

Projection:
project

Slice:
slice

Volume render:
render

### DIFF
--- a/swiftsimio/visualisation/projection.py
+++ b/swiftsimio/visualisation/projection.py
@@ -18,6 +18,7 @@ from numpy import (
     empty_like,
     logical_and,
     s_,
+    append,
 )
 from unyt import unyt_array, unyt_quantity, exceptions
 from swiftsimio import SWIFTDataset, cosmo_array
@@ -231,13 +232,32 @@ def project_pixel_grid(
     else:
         combined_mask = mask
 
-    common_arguments = dict(
-        x=(x[combined_mask] - x_min) / max_range,
-        y=(y[combined_mask] - y_min) / max_range,
-        m=m[combined_mask],
-        h=hsml[combined_mask] / max_range,
-        res=resolution,
-    )
+    xfinal = array((x[combined_mask] - x_min) / max_range)
+    yfinal = array((y[combined_mask] - y_min) / max_range)
+    mfinal = array(m[combined_mask])
+    hfinal = array(hsml[combined_mask] / max_range)
+    rescaled_box = array([box_x / max_range, box_y / max_range])
+
+    xall = array([])
+    yall = array([])
+    mall = array([])
+    hall = array([])
+    for xshift in [-1, 0, 1]:
+        for yshift in [-1, 0, 1]:
+            thisx = xfinal + xshift * rescaled_box[0]
+            thisy = yfinal + yshift * rescaled_box[1]
+            inside = (
+                (thisx - xshift * hfinal <= rescaled_box[0])
+                & (thisx - xshift * hfinal >= 0.0)
+                & (thisy - yshift * hfinal <= rescaled_box[1])
+                & (thisy - yshift * hfinal >= 0.0)
+            )
+            xall = append(xall, thisx[inside])
+            yall = append(yall, thisy[inside])
+            mall = append(mall, mfinal[inside])
+            hall = append(hall, hfinal[inside])
+
+    common_arguments = dict(x=xall, y=yall, m=mall, h=hall, res=resolution)
 
     if parallel:
         image = backends_parallel[backend](**common_arguments)
--- a/swiftsimio/visualisation/projection.py
+++ b/swiftsimio/visualisation/projection.py
@@ -238,12 +238,14 @@ def project_pixel_grid(
     hfinal = array(hsml[combined_mask] / max_range)
     rescaled_box = array([box_x / max_range, box_y / max_range])
 
-    xall = array([])
-    yall = array([])
-    mall = array([])
-    hall = array([])
+    xall = xfinal.copy()
+    yall = yfinal.copy()
+    mall = mfinal.copy()
+    hall = hfinal.copy()
     for xshift in [-1, 0, 1]:
         for yshift in [-1, 0, 1]:
+            if xshift == 0 and yshift == 0:
+                continue
             thisx = xfinal + xshift * rescaled_box[0]
             thisy = yfinal + yshift * rescaled_box[1]
             inside = (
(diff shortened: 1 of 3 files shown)


JBorrow:
Excellent change Bert, thanks. I will review the code later today.

JBorrow:
This change looks like it may introduce significant performance impacts. Is there some way of checking this?

JBorrow:
I would also suggest it may be quicker to just run the vis code twice (one for periodic particles, one for non-periodic) than append to the arrays.

bwvdnbro:
Of course there will be a performance impact. We are including more particles.

Some quick timings based on the scripts I used for the images above (times are the mean and standard deviation for 10 successive calls to the corresponding visualisation function):
```
master:
project:
Took 10732.50 +- 230.36167937716385 ms
slice:
Took 357.06 +- 198.60511500139805 ms
render:
Took 832.70 +- 227.8507330442371 ms

new branch:
project:
Took 12704.47 +- 222.1358350366033 ms
slice:
Took 392.93 +- 181.07175431095246 ms
render:
Took 1062.95 +- 210.43395451782126 ms
```
So roughly ~20% more.

I don't think appending to the arrays is particularly slow, because this only happens 9 or 27 times and scales linearly with the number of particles. The runtime is generally dominated by the scatter function. I could eliminate the iteration of the loops that deals with the normal case (`x/y/zshift==0`), but that would make the code uglier for probably little gain.
Calling the scatter function twice has the disadvantage that it introduces an extra synchronisation point if you are using a parallel scatter function. So you increase the potential for dead time because of load imbalances.
The only good reason I can see to call the scatter function twice (or even more) is to reduce the memory footprint. That will be larger in the current implementation and is generally unpredictable (worst case, it could be 27 times larger for the volume rendering). But I doubt this will be a serious problem for any real application.

JBorrow:
Does it introduce another sync point? The addition to the main array, as far as I know, is actually handled asynchronously.

JBorrow:
You can also do it in parallel, add scatter_inside + scatter_outside then add this to the main array?

JBorrow:
Hmm, looking at the code I see what you mean about sync point. But this should be well load-balanced, anyway.

bwvdnbro:
> Does it introduce another sync point? The addition to the main array, as far as I know, is actually handled asynchronously.

I might be misunderstanding what you meant then. I thought you were suggesting to have multiple calls to
```
    if parallel:
        image = backends_parallel[backend](**common_arguments)
    else:
        image = backends[backend](**common_arguments)
```
This is not done in any parallel bit of the code, so at the end of each call you would get a new synchronisation point.

Anyway, I have tested this and is definitely slower than calling the backend once with the appended arrays.

What does seem to make a difference is initialising the arrays with the particles and skipping the `x/y/zshift==0` case in the loop. At least for the projection. For the slice and volume rendering this is actually slower and I don't understand why.

JBorrow:
I would like to see this trailed on something a bit bigger before merging.

In particular, I'm concerned about memory consumption for the volume rendering, as this is already extremely memory-intensive (though that mainly comes from allocating the voxel grid).

swiftsimio/visualisation/projection.py
```diff
@@ -231,13 +232,34 @@ def project_pixel_grid(
...
+        for yshift in [-1, 0, 1]:
+            if xshift == 0 and yshift == 0:
+                continue
+            thisx = xfinal + xshift * rescaled_box[0]
+            thisy = yfinal + yshift * rescaled_box[1]
+            inside = (
+                (thisx - xshift * hfinal <= rescaled_box[0])
+                & (thisx - xshift * hfinal >= 0.0)
+                & (thisy - yshift * hfinal <= rescaled_box[1])
+                & (thisy - yshift * hfinal >= 0.0)
```
JBorrow, line 255: dx = thisx - xshift * hfinal
dy = thisy - yshift * hfinal

Is np.logical_and.reduce() faster?

swiftsimio/visualisation/projection.py
```diff
@@ -231,13 +232,34 @@ def project_pixel_grid(
...
+    xfinal = array((x[combined_mask] - x_min) / max_range)
+    yfinal = array((y[combined_mask] - y_min) / max_range)
+    mfinal = array(m[combined_mask])
+    hfinal = array(hsml[combined_mask] / max_range)
+    rescaled_box = array([box_x / max_range, box_y / max_range])
+
+    xall = xfinal.copy()
+    yall = yfinal.copy()
+    mall = mfinal.copy()
+    hall = hfinal.copy()
```
JBorrow, line 244: Is there some way that we can work out if we're going to see the edges ahead of time..? Just thinking that we can skip this step.

swiftsimio/visualisation/projection.py
bwvdnbro: Yes, I should precompute dx and dy.

According to the line profiler, `logical_and.reduce()` is actually slower.

swiftsimio/visualisation/projection.py
bwvdnbro: Not without knowing what the maximum smoothing length inside the visualisation region is. And that would require an extra calculation that is also not cheap.

According to the line profiler, this bit of code takes only ~5% of the total time in the function, so the gains are small.

bwvdnbro:
I have run the projection code on a thin slice through one of the 1 Gpc FLAMINGO snapshots and the results are not great. The memory footprint and run time increase quite dramatically. I'll try some more things.

swiftsimio/visualisation/projection.py
JBorrow: What about (x & x) & (y & y)? Or is this not faster... Maybe python is smart enough to reduce like that

JBorrow:
Is it worth changing the numba functions to check if the particle overlaps the edge, then seeing if it would partially project onto the other side? This way you would only have one copy of data in memory, and branch prediction would likely take care of any negative performance impacts?

bwvdnbro:
I was hoping to avoid touching the numba functions, so that this is implemented in one place instead of in every single backend. But I guess that will indeed be a better solution given the impact on the performance.

I would like to keep this branch alive for reference until I get a chance to properly implement this in the backends. Having a slow but working version is useful for testing.

MatthieuSchaller:
Given the other branch was merged, can we close this one?

JBorrow:
Yes, sorry for leaving this open.

PR #150, opened 2022-09-14
Made sure visualisation functions include periodic copies of particles (attempt 2)

bwvdnbro:
New version of #148 that implements the periodic wrapping on the backend level. The memory footprint is not affected by this version, but there should still be a (small) performance penalty.

Here are some timings for my small test based on `cosmological_volume.hdf5`:

Old version:
```
Projection:
Backend "histogram" took 743.87 +- 12.995156135586896 ms
Backend "fast" took 620.74 +- 32.442475264777684 ms
Backend "renormalised" took 1128.18 +- 201.63854249894925 ms
Backend "subsampled" took 2442.19 +- 135.6247984160516 ms
Backend "subsampled_extreme" took 2361.57 +- 158.01341649057622 ms
Backend "reference" took 1250.00 +- 166.40818857369823 ms
Backend "gpu" took 41.47 +- 10.81907894761692 ms
Slicing:
Took 419.78 +- 8.391963275394627 ms
Volume rendering:
Took 4440.49 +- 59.48874302494967 ms
```

New version:
```
Projection:
Backend "histogram" took 733.99 +- 33.32694376915969 ms
Backend "fast" took 584.81 +- 20.463084650768796 ms
Backend "renormalised" took 785.32 +- 20.98768985286519 ms
Backend "subsampled" took 1270.11 +- 28.538669311331958 ms
Backend "subsampled_extreme" took 1263.11 +- 44.949929481954534 ms
Backend "reference" took 756.16 +- 27.871183567513178 ms
Backend "gpu" took 53.32 +- 13.414277270534988 ms
Slicing:
Took 411.87 +- 8.084733606619707 ms
Volume rendering:
Took 4338.15 +- 241.85104295403843 ms
```

For some reason all of them (except for the gpu backend) were actually faster in the new version. Not sure I understand why.

I am still running a larger and more representative test.

### DIFF
--- a/docs/source/visualisation/index.rst
+++ b/docs/source/visualisation/index.rst
@@ -17,6 +17,7 @@ common interface:
        project="masses", # Variable to project, e.g. masses, temperatures, etc.
        parallel=False, # Construct the image in (thread) parallel?
        region=None, # None, or a list telling which region to render_func_name
+       periodic=True, # Whether or not to apply periodic boundary conditions
    )
 
 The output of these functions comes with associated units and has the correct
--- a/docs/source/visualisation/projection.rst
+++ b/docs/source/visualisation/projection.rst
@@ -30,7 +30,13 @@ Example
 
    # This creates a grid that has units msun / Mpc^2, and can be transformed like
    # any other unyt quantity
-   mass_map = project_gas(data, resolution=1024, project="masses", parallel=True)
+   mass_map = project_gas(
+       data,
+       resolution=1024,
+       project="masses",
+       parallel=True,
+       periodic=True,
+   )
 
    # Let's say we wish to save it as msun / kpc^2,
    from unyt import msun, kpc
@@ -61,13 +67,20 @@ this:
    data.gas.mass_weighted_temps = data.gas.masses * data.gas.temperatures
 
    # Map in msun / mpc^2
-   mass_map = project_gas(data, resolution=1024, project="masses", parallel=True)
+   mass_map = project_gas(
+       data,
+       resolution=1024,
+       project="masses",
+       parallel=True,
+       periodic=True,
+   )
    # Map in msun * K / mpc^2
    mass_weighted_temp_map = project_gas(
        data,
        resolution=1024,
        project="mass_weighted_temps",
-       parallel=True
+       parallel=True,
+       periodic=True,
    )
 
    temp_map = mass_weighted_temp_map / mass_map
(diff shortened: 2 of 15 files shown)


bwvdnbro:
I have tested this on a 20 Mpc thick slab of a 1Gpc FLAMINGO simulation. All backends except for `reference` (which spits out lots of "Reference grid not created at a high enough resolution" messages and actually gives the wrong result) are faster in the new version, despite doing a bit more work. The results are the same as with the old version.

JBorrow:
Reference should do that, it's designed to make sure that every particle's kernel is well resolved (if you reach a particle that is not resolved, it quits - hence the wrong answer).

The other stuff looks great, thank you. I am concerned that we are changing an API that is used by external code here, though. Is there some way that we can make box_x and box_y optional? Is there a performance impact if they are set to None in the case where you don't want periodicity?

bwvdnbro:
I guess that would be possible, yes. I will check that.

bwvdnbro:
Numba does not look `Union[float64, None]`, so I have made the default `box_x/y=0.`, which actually makes sense as well. I have now also added an optional argument `periodic = True` to all visualisation functions. The non-periodic version is slightly faster, as expected.

Do you think it would be worth adding something about periodic boundaries to the documentation? I had a quick look, but I feel that this does not really fit well within the current text.

JBorrow:
I don't think we need to describe the general concept of periodic boundaries (we should leave that to the SWIFT docs) but we'll need to update the docs with the new function arguments and describing the general concept of projecting over periodic boundaries.

bwvdnbro:
I have updated the documentation.

JBorrow:
Final question: does this use the periodic flag to see if the projection should be done periodically?

bwvdnbro:
No, it does not. It also took me a while to find a "periodic flag". The only place where the periodic flag is stored is in the parameter list in the snapshot, as `InitialConditions:periodic`. I could try to read that, but that is not very clean. Is there another "periodic flag" I am not aware of?

bwvdnbro:
@JBorrow any thoughts about my last comment?

The default behaviour now is to assume periodic boundaries if not specified, which seems sensible for most actual applications. The default for the backends themselves is to assume no periodic boundaries, since those do not automatically have access to the box size.

JBorrow:
Okay, that's fine then. Let's keep it that way, but make sure to document this behavior choice thoroughly.

Have you checked for correctness? I suppose there are the automated tests, but are there periodic correctness tests?

bwvdnbro:
I have done the tests described in the previous version of this PR (#148). Not sure if there is something we can do automatically to test this.

JBorrow:
Is it possible to construct one large kernel that overlaps the boundary and check that the answer is the same in periodic mode and adding two non-periodic images together?

E.g. something like this:

```
-          -
---      ---
----    ----
---      ---
-          -
```

PR #152, opened 2022-11-04
Added neutrinos and Tcmb0 read from snapshots

JBorrow:
Fixes #151.

Can @JegerBroxterman please test this and make sure that it is consistent with their expectations?

PR #154, opened 2022-12-13
Convert SWIFT neutrino mass and degeneracy format to astropy format

wullm:
This should work for all the FLAMINGO runs and most reasonable models.


wullm:
Follow-up on #152

Issue #153, opened 2022-12-06
Too much opening and closing files


bwvdnbro:
Last week I got told off by Peter Draper for overloading the meta-data server of the cosma7 file system. The reason: I was running 192 parallel instances of a script that uses swiftsimio to extract galaxies from a snapshot for analysis. Apparently the meta-data server is queried whenever a file is opened, even if you just closed that same file after another reading operation. In swiftsimio, the snapshot file is opened (and almost immediately closed again):
 - when reading header information
 - when reading the list of datasets
 - when reading units meta-data
 - when reading basically every bit of meta-data for a dataset (the datasets are done in one read, but a separate read is used for different bits of meta-data).
 - when reading the cosmology
 - when reading the named columns
 - when reading individual datasets
That is a lot. I understand the reason for reading datasets separately, but I don't see a good reason to open and close the file for meta-data; this can and should all be read in one go. That might already help a bit.

Even if the number of reads per `load()` can be reduced, the more fundamental issue is that, right now, we need to do a separate `load()` whenever the mask is changed (because you want to read a different galaxy for example). That is great for parallelisation, since it essentially means you can spawn lots of processes for different galaxies without needing to worry about parallel HDF5 (each process uses its own `h5py` instance). But it means a huge overhead (in file accesses and also in amount of data being read) when you are trying to read a significant fraction of the snapshot. We need a better mechanism to deal with that. Ideally, we would have a mechanism that reads large chunks of datasets in few operations (Lustre file systems like that), but hides this from the user.

@JBorrow any thoughts on this? We really want to use swiftsimio for the morphology pipeline we plan to use for COLIBRE, but without some way to improve performance for loops over galaxies, we will never be able to run this on a large box.

JBorrow:
We could keep around a persistent handle for the snapshot and re-use it. I've avoided doing that so far as it seemed cleaner to me this way, and you could ask the question of how on earth we would ever give back the handle. I guess we would have that in the destructor of the swift dataset, but you would still need to query the metadata server once per galaxy (though that doesn't seem too bad to me).

JBorrow:
From slack: Is this not a massive concern for distributed snapshots? There is surely no other way of doing this than swapping out file handles, otherwise we'll have thousands of handles hanging around. Maybe @jchelly has some insight here? Does HDF5 do some magic under the covers?

JBorrow:
[...]

PR #155, opened 2022-12-16
Allows file handles to stick around longer

JBorrow:
Attempted remedy to #153.

This stores a file handle in SWIFTUnits, which is accessed by SWIFTMetadata. It is accessed through a property, meaning that the file can be re-opened if necessary.

@bwvdnbro Is it possible to test this (with peter's blessing?)

If we need to go further, we can stick the file handle into the dataset reading. But those should take long enough that grabbing file handles is ok.

### DIFF
--- a/swiftsimio/masks.py
+++ b/swiftsimio/masks.py
@@ -75,50 +75,47 @@ class SWIFTMask(object):
         self.counts = {}
         self.offsets = {}
 
-        with h5py.File(self.metadata.filename, "r") as handle:
-            cell_handle = handle["Cells"]
-            count_handle = cell_handle["Counts"]
-            metadata_handle = cell_handle["Meta-data"]
-            centers_handle = cell_handle["Centres"]
+        cell_handle = self.units.handle["Cells"]
+        count_handle = cell_handle["Counts"]
+        metadata_handle = cell_handle["Meta-data"]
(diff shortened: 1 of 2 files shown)


bwvdnbro:
@JBorrow I'm not sure what I would test here. I can check that this works, but I don't really have a good way of confirming if this helps with the unusually large load on the meta-data server. Peter telling me off is not exactly a very good diagnostic. I also expect that it will be hard to get hold of 192 nodes at the same time again. One would hope that the queue is usually busy enough to prevent that.

I don't think it can do any harm to open the file less though.

JBorrow:
Would it be possible to run the morphology pipeline to make sure nothing is messed up?

MatthieuSchaller:
What is the status of this? Do we still want the change? If so what can we do to test it?

JBorrow:
I would like to see somebody run this at scale and see if it breaks anything. It is a generally good change, and does not break anything. For that reason we can probably just merge it in and see if anyone complains...

MatthieuSchaller:
@jchelly or @kyleaoman may have large test cases.

kyleaoman:
I don't think that I do...

JBorrow:
Well it's not destructive and just seems to be a positive change, so I am merging it.

PR #157, opened 2023-03-08
Make the reading of strings in snapshots header more robust to h5py's…

MatthieuSchaller:
Make the reading of strings in snapshots header more robust to h5py's inconsistent behaviour vis-a-vis encoding of fixed- and variable-length arrays

Fixes a problem encountered in FLAMINGO where hand-edited (via h5py) names in snapshots lead to us not being able to open the files.

### DIFF
--- a/swiftsimio/reader.py
+++ b/swiftsimio/reader.py
@@ -413,7 +413,7 @@ class SWIFTMetadata(object):
                 raw = self.header[field]
                 try:
                     string = raw.decode("utf-8")
-                except:
+                except AttributeError:
                     string = raw
                 setattr(self, name, string)
             except KeyError:
--- a/swiftsimio/reader.py
+++ b/swiftsimio/reader.py
@@ -407,7 +407,15 @@ class SWIFTMetadata(object):
 
         for field, name in metadata.metadata_fields.header_unpack_string.items():
             try:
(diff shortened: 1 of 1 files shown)


swiftsimio/reader.py
```diff
@@ -407,7 +407,15 @@ def postprocess_header(self):
...
         for field, name in metadata.metadata_fields.header_unpack_string.items():
             try:
-                setattr(self, name, self.header[field].decode("utf-8"))
+                # Deal with h5py's quirkiness that fixed-sized and variable-sized
+                # strings are read as strings or bytes
+                # See: https://github.com/h5py/h5py/issues/2172
+                raw = self.header[field]
+                try:
+                    string = raw.decode("utf-8")
+                except:
```
jchelly, line 416: Probably best not to use a bare except. Apparently that even catches ctrl-c! Maybe we can just catch the AttributeError when it has no decode() method, or use isinstance to see if we got str or bytes?

MatthieuSchaller: Ah! I learned something about exceptions. :)

Thanks!

JBorrow:
Yep, UNIX signals are handled by exceptions within python itself, so fun! Thanks for this Matthieu, I will push a new release.

PR #159, opened 2023-04-04
Added new superclass for unyt3.0 behaviour

JBorrow:
Unyt3 removes the deprecated `input_units` argument.


JBorrow:
Yes, probably. I don't know how SOAP is implemented @jchelly.

PR #160, opened 2023-06-29
Move from setup.py to pyproject.toml

kyleaoman:
It seems like recent versions of python (I'm on 3.11 now) are ignoring setup.py and when they fail to find a pyproject.toml file, they fall back to guesswork and then fall over. I've had a go at migrating to a pyproject.toml build setup. It seems to be working, but:

- I haven't added in the long description that was in the setup.py file (programmatically retrieved from the README), not sure how to replicate that in this framework.
- The version is hard-coded in the pyproject.toml file. I did some googling to see if it could be linked to the __version__.py file, and it seems like the answer is "we won't support that, please don't do it".
-  Currently tests fall over because pysphviewer doesn't install... because it also needs to move from setup.py to pyproject.toml (I think). Will try to make a PR for that soon, too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
 dependencies = [
     "numpy",
     "h5py",
-    "unyt>=3.0.0",
+    "unyt>=2.9.0",
     "numba>=0.50.0",
 ]
 


kyleaoman:
Actually with all the C extensions to be built, converting py-sphviewer from setup.py to pyproject.toml is way above my skill level. Given that Ale doesn't maintain the package anymore, is it really needed as a dependency?

kyleaoman:
Tentatively chopped out everything swiftsimio related from the source code. Occurs to me now that it would also have to be removed from the docs.

kyleaoman:
Also removed pysphviewer from the docs. Silently purging all of this is perhaps too drastic, I could go back and add some things back with deprecation warnings if you want - the main thing is that pysphviewer won't install during CI currently, so either someone has to figure out a pyproject.toml for that package, or we drop it from the unit tests (and possibly the list of optional dependencies).

pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+version="6.1.1"
+authors = [
+    { name="Josh Borrow", email="joshua.borrow@durham.ac.uk" },
+]
+description="SWIFTsim (swift.dur.ac.uk) i/o routines for python."
+readme = "README.md"
+requires-python = ">3.8.0"
+classifiers = [
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
```
JBorrow, line 19: Need to add the 3.10 and 3.11 given we changed to support 3.11 :)

pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+    "License :: OSI Approved :: GNU Lesser General Public License v3 or later (LGPLv3+)",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+    "numpy",
+    "h5py",
+    "unyt>=2.9.0",
```
JBorrow, line 26: Do we want to show `unyt>=3.0.0` now?

JBorrow:
Minor things then happy to merge.

pyproject.toml
kyleaoman: Actually I don't see unyt >=3.0.0 on PyPI even on python 3.11...

kyleaoman:
@JBorrow minor things addressed, comments inline above.

pyproject.toml
JBorrow: Ah right unyt3.0.0 has not yet been released. My bad!

Issue #169, opened 2023-09-28
Opening one of the sub-file of a distributed snaphot leads to unexpected error when using the cell masking


MatthieuSchaller:
Bug reported on Slack by @Joeybraspenning.

Code snippet running on a flamingo box and reading the `.0` file rather than the virtual file:

```
    load_region = [[xc - 2*r500c, xc+2*r500c], [yc-2*r500c, yc+2*r500c], [zc-2*r500c, zc+2*r500c]]
    print(f'{load_region=}')
    # Constrain the mask
    mask.constrain_spatial(load_region)
    # Now load the snapshot with this mask
    data = sw.load(filename, mask=mask)
    print(f'{np.min(data.gas.coordinates, axis = 0)=}')
    print(f'{np.max(data.gas.coordinates, axis = 0)=}')
 ```
 leads to 
```
load_region=[[unyt_quantity(102.72425057, 'Mpc'), unyt_quantity(104.15979745, 'Mpc')], [unyt_quantity(89.56892057, 'Mpc'), unyt_quantity(91.00446745, 'Mpc')], [unyt_quantity(3.98413057, 'Mpc'), unyt_quantity(5.41967745, 'Mpc')]]
np.min(data.gas.coordinates, axis = 0)=cosmo_array([125.00008293, 687.50001293, 343.75001293], 'Mpc')
np.max(data.gas.coordinates, axis = 0)=cosmo_array([156.24988293, 718.74998293, 406.24991293], 'Mpc')
```

The code should either use the meta-data to open the correct sub-file or throw an error about trying to access
a region not present in the file.

**Comment by Joeybraspenning** [comment]
This might be the better code snippet to refer to as it includes the read of the file

```
import swiftsimio as sw
import numpy as np
from unyt import Mpc

filename = "/cosma8/data/dp004/flamingo/Runs/L1000N1800/HYDRO_FIDUCIAL/snapshots/flamingo_0077/flamingo_0077.0.hdf5"
mask = sw.mask(filename)

load_region = [[10*Mpc, 15*Mpc], [100*Mpc, 110*Mpc], [200*Mpc, 210*Mpc]]
print(load_region)
# Constrain the mask
mask.constrain_spatial(load_region)
# Now load the snapshot with this mask
data = sw.load(filename, mask=mask)

print(np.min(data.gas.coordinates, axis = 0))
print(np.max(data.gas.coordinates, axis = 0))
```

JBorrow:
[...]

PR #170, opened 2023-10-06
Add exception when trying to load a partial snapshot with partial loading

JBorrow:
"Fixes" #169 by notifying the user if they are trying to do something naughty.

@Joeybraspenning can you try this branch out and let me know if it crashes with your old use case?


MatthieuSchaller:
@Joeybraspenning this should fix the issue you reported. Or at least tell you there is an error.

Joeybraspenning:
Ah sorry, missed this, will try within 24h

MatthieuSchaller:
@Joeybraspenning any updates?

JBorrow:
@Joeybraspenning a reminder we're waiting for you to test this before we can merge

Joeybraspenning:
Having a look now, github is connected to my personal email which I tend to ignore, sorry about that

Joeybraspenning:
Seems all good now, I get an error telling me I can't create masks on partial snapshots:

```
swiftsimio.objects.InvalidSnapshot: InvalidSnapshot: You cannot use masks on partial snapshots. Please use the virtual file generated by SWIFT (use snapshot.hdf5, not snapshot.0.hdf5).
```

That's sufficient for me, prevents me from ever making this mistake by accident

PR #176, opened 2023-10-23
Add API functionality

harryjmoss:
As part of work between UCL and Durham, we have produced a prototype FastAPI app for sending numpy arrays derived from HDF5 files, as well as other SWIFT objects.

[Here is the API](https://github.com/UCL-ARC/dirac-swift-api)


kyleaoman:
@JBorrow the time allocated to the team at UCL to work on this has now ended so they're handing what they have (which is now fairly mature) over to us. I think that what we should do from here is:

1. Merge this PR into a feature branch to wrap up some loose ends before merging to master.
2. In that feature branch move the API code into a separate module (or submodule?) of swiftsimio. I think it makes sense to distribute it in the swiftsimio package.
3. After that there's a little bit of documentation work to do in the swiftsimio narrative docs. At a minimum we should point to the remote API docs as an available feature, and probably add a new page with some cookbook examples. This would be about the earliest to consider merging into master?
4. Then there's a bit of work to do to set up a persistent server on cosma (on dataweb, I think). The RSEs have left some suggestions of plausible ways this could work.
5. A little bit of development needed to decide how we want to offer available simulation data through the server. Loosely the server would maintain a dictionary mapping labels to hdf5 files on disc, something like `{"flamingos_NxxxxLxxxx_physicslabel_snapnum": "/cosma8/.../snap.hdf5"}`, and then a user should be able to set up a `RemoteSWIFTDataset` with a label plus their credentials.
6. Set up a webpage (on/linked from the VirgoDB pages?) pointing to the relevant documentation and tabulating the data available from the server by label. Maybe we can set everything up with a test dataset and then roll out an initial simulation dataset that we can announce more widely. Perhaps Flamingos makes a good candidate?

pyproject.toml
```diff
@@ -23,10 +23,12 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
+    "cloudpickle>=2.2.1",
     "numpy",
     "h5py",
     "unyt>=2.9.0",
     "numba>=0.50.0",
+	"requests>=2.31.0",
```
JBorrow, line 31: Weird indentation

swiftsimio/reader.py
```diff
@@ -16,16 +16,28 @@
...
 
 from typing import Union, Callable, List, Optional
+from pathlib import Path
+
+
+class NumpyEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        return json.JSONEncoder.default(self, obj)
```
JBorrow, line 40: In general it would be great to have all the remote stuff in a separate file (e.g. remote.py)

swiftsimio/reader.py
```diff
@@ -232,6 +244,21 @@ def __del__(self):
...
+                if key not in excluded_fields and not isinstance(
+                    value, unyt.unyt_quantity
+                ):
+                    if isinstance(unit_dict[key], dict):
+                        for nested_key, nested_value in unit_dict[key].items():
+                            setattr(self, nested_key, nested_value)
+                    else:
+                        setattr(self, key, unyt.unyt_quantity.from_string(value))
+
+
```
JBorrow, line 261: What on earth does this do? Why does it exist? Needs to be documented.

swiftsimio/reader.py
```diff
@@ -1196,6 +1223,198 @@ def getter(self):
     return getter
 
 
+def load_ndarray_from_json(json_array: str, data_type: Union[str, None]) -> npt.NDArray:
+    """Convert JSON to a Numpy NDArray.
+
+    Args:
+        json_array (str): Numpy array as JSON
+        data_type (str): Data type of elements in the original array
```
JBorrow, line 1231: Documentation is inconsistent with the rest of the codebase.

swiftsimio/reader.py
```diff
@@ -1196,6 +1223,198 @@ def getter(self):
...
+    -------
+        npt.NDArray: Numpy NDArray object
+    """
+    loaded_json = json.loads(json_array)
+    return np.asarray(loaded_json, dtype=data_type)
+
+
+def generate_getter_remote(
+    server_address: str,
+    session: requests.Session,
```
JBorrow, line 1243: Session is undocumented

swiftsimio/reader.py
```diff
@@ -1612,3 +1979,222 @@ def create_particle_datasets(self):
...
+    pass
+
+
+class RemoteSWIFTDataset:
+    """
+    A collection object for:
+
+    + SWIFTUnits,
+    + SWIFTMetadata,
+    + SWIFTParticleDataset
```
JBorrow, line 1998: Seems to be for the remote versions of these.

JBorrow (commented):
Some minor comments; in particular I would like to see these moved outside of this file.

Is it possible to test these functions? I would like to see that.

kyleaoman:
Thanks @JBorrow. It's on my list to make a draft PR for this branch with a to-do list before it can be merged. Getting testing set up alone is going to be a big job. Ideally a lot of existing tests should be refactored to use fixtures so that we can test both the stand-alone and "client" SWIFTDatasets. Always hesitant to start making significant changes to tests, though, so will want to brainstorm that a bit before investing any effort.  Figuring out how to host a server for the tests is also a 
bit of a tricky one - probably possible but not similar to anything I've tried before. Lots of other little stuff to do, too.

PR #182, opened 2024-02-22
New slicing backend (nearest neighbors interpolation) for use with moving mesh

yuyttenhove:
I implemented a new slicing backend (similar to how there are multiple backend for the projection) which uses nearest neighbor interpolation of the quantities as this is more suitable for moving mesh hydrodynamics.

All tests pass and I modified the relevant ones to also test the new backend. All features of the original slicing method (now still the default and dubbed `"sph"`) should be supported by the new `"nearest_neighbors"` backend. No default behaviour has changed, so this should be fully backwards compatible.

I also added a few sentences to the documentation.

Do you think this is suitable for including in `swiftsimIO`? 
Do you have any remarks/wishes?


yuyttenhove:
As a simple comparison of the 2 backends:

`"sph"` | `"nearest_neighbors"`
:---:|:---:
densities_sph | densities_nearest_neighbors

JBorrow:
Amazing! Love it. Thank you for this contribution; we should definitely include this. Thank you also for so carefully working within the existing framework.

swiftsimio/visualisation/slice.py
```diff
@@ -387,6 +80,10 @@ def slice_gas_pixel_grid(
         Particles outside of this range are still considered if their
         smoothing lengths overlap with the range.
 
+    backend : str, optional
+        Backend to use. Choices are "sph" (default) for interpolation using kernel weights or
+        "nearest_neighbours" for nearest neighbour interpolation.
+
     periodic : bool, optional
```
JBorrow, line 87: Please make sure to have comment line lengths be consistent with code. I like to keep them under 80.

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
...
+from swiftsimio.visualisation.slice_backends.sph import get_hsml as sph_get_hsml
+
+
+def _build_tree(
+    x: float64, y: float64, z: float64, box_x: float64, box_y: float64, box_z: float64
+) -> KDTree:
+    """
+    Build the tree used for the nearest-neighbor calculations.
+    In the periodic case, we must make sure that all particle coordinates fall inside the box.
+
```
JBorrow, line 14: Same comment here.

I also do not like 'private' method syntax in sub-files like this. Please just call this 'build_tree'

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
...
+        Used for periodic wrapping.
+
+    Returns
+    -------
+    KDTree object
+        A KD-tree built from the particle positions
+    """
+    if not TREE_AVAILABLE:
+        raise ImportError(
+            'The scipy.spatial.cKDTree class is required to use the "nearest_neighbors" slice backend.'
```
JBorrow, line 40: String too long, also do not use double quotes inside as that is different to black style (feel free to change this to "The scipy.spatial.cKDTree class is required to use the 'nearest_neighbors' slice backend.")

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
+from numpy import float64, float32, ndarray, linspace, array, stack, meshgrid, power
+
+from swiftsimio import SWIFTDataset, cosmo_array
+from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
+from swiftsimio.visualisation.slice_backends.sph import get_hsml as sph_get_hsml
+
+
+def _build_tree(
+    x: float64, y: float64, z: float64, box_x: float64, box_y: float64, box_z: float64
```
JBorrow, line 9: Consider `float` not `float64` as these are not strictly numpy types (could also even be float32s)

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
...
+    z: float64,
+    m: float32,
+    h: float32,
+    z_slice: float64,
+    xres: int,
+    yres: int,
+    box_x: float64 = 0.0,
+    box_y: float64 = 0.0,
+    box_z: float64 = 0.0,
+    workers: int = 1,
```
JBorrow, line 68: Regular floats may be better here again. Also consider None for default values where there is diffferent behaviour (e.g. box_x)

JBorrow (commented):
Some minor comments, but looks good.

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
yuyttenhove: You mean only for the `z_slice` and `box_x/y/z` paramters, right? (Since `x`, `y` and `z` are numpy arrays)

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
...
+    h: float32,
+    z_slice: float64,
+    xres: int,
+    yres: int,
+    box_x: float64 = 0.0,
+    box_y: float64 = 0.0,
+    box_z: float64 = 0.0,
+) -> ndarray:
+    """
+    Creates a scatter plot of the given quantities for a particles in a data slice.
```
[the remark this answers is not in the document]
yuyttenhove: Yeah, I just copied that verbatim from the other implementation. Should I also change it there?

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,274 @@
...
+        raise ImportError(
+            'The scipy.spatial.cKDTree class is required to use the "nearest_neighbors" slice backend.'
+        )
+    if box_x != 0 or box_y != 0 or box_z != 0:
+        if box_x != 0:
+            x[x < 0] += box_x
+        if box_y != 0:
+            y[y < 0] += box_y
+        if box_z != 0:
+            z[z < 0] += box_z
```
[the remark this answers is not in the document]
yuyttenhove: I did it like this, because `box_x/y/z` are computed higher up (in `slice_gas_pixel_grid`) and are set to `0.0` if periodic is `false`. The original implementations also rely on them being numeric, so I cannot just change that whithout also changing the `"sph"` backend. 

I could just provide no default values?

yuyttenhove:
Thanks for the comments! I resolved most of them and left some questions.

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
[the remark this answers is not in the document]
JBorrow: Yes, that's worth doing. Thank you.

swiftsimio/visualisation/slice_backends/nearest_neighbors.py
```diff
@@ -0,0 +1,236 @@
+from numpy import float64, float32, ndarray, linspace, array, stack, meshgrid, cbrt
+
+from swiftsimio import SWIFTDataset, cosmo_array
+from swiftsimio.optional_packages import KDTree, TREE_AVAILABLE
+from swiftsimio.visualisation.slice_backends.sph import get_hsml as sph_get_hsml
```
JBorrow, line 5: My final comment is that we should probably move get_hsml to a higher level? As you will likely want to do projections too.

You could add this at the swiftsimio.visualisation level?

yuyttenhove: Yes, good idea

yuyttenhove:
That should be it then, I think?

JBorrow:
Looks good. We will need to release a new version (v8.0.0) as this contains api breaking changes.

yuyttenhove:
I just pushed another commit in which I adjusted the spelling of neighbors to neighbours (like it was already used in `generate_smoothing_lengths`)

yuyttenhove:
I turns out np.cbrt is actually not supported by `unyt_array`s (and hence `cosmo_array`s). I opened a pull request to add support for it in unyt and changed the `get_hsml` implementation in the meantime.

JBorrow:
Not required for this PR, but you could imagine actually not even needing a tree for this. For densely populated areas, you can just find the particle nearest to the pixel center by looping through all the particles, and for the sparsely populated areas a flood fill after that first procedure would work...

yuyttenhove:
> Not required for this PR, but you could imagine actually not even needing a tree for this. For densely populated areas, you can just find the particle nearest to the pixel center by looping through all the particles, and for the sparsely populated areas a flood fill after that first procedure would work...

Yeah, I did consider that actually, as it will probably be faster, but ended up using the tree method, since I already used it outside swiftwimio and hence was very easy to implement. Something to keep in mind for later probably though.

JBorrow:
Thank you so much for this contribution!

PR #190, opened 2024-06-18
Load FOF catalogues & Cosmology Metadata Changes

robjmcgibbon:
Allow swiftsimio to load FOF files. I've made the `SWIFTDataset` class more generic, so it now has a `filetype` attribute (currently can be snapshot or FOF). Instead of storing an int for each particle type it now saves the names of the groups within the hdf5 that we want to load ('PartType{i}' for snapshots, and 'Groups' for fof files).

This seems to work for loading both snapshots and fof files. I'll have a look at the tests next.


swiftsimio/metadata/particle/particle_types.py
```diff
@@ -4,31 +4,31 @@
...
-    4: "Stars",
-    5: "Black Holes",
-    6: "Neutrinos",
+    "PartType0": "Gas",
+    "PartType1": "Dark Matter",
+    "PartType2": "Boundary",
+    "PartType3": "Sinks",
+    "PartType4": "Stars",
+    "PartType5": "Black Holes",
+    "PartType6": "Neutrinos",
```
JBorrow, line 33: Why did you change this?

robjmcgibbon: I changed `present_particle_types` to return "PartType{i}" rather than integers
https://github.com/SWIFTSIM/swiftsimio/blob/143d65ff3b68cf85f6aba914885b0e96295ed769/swiftsimio/reader.py#L652
So I could either change this file, or else I'd have to convert back to an integer when getting the particle name

robjmcgibbon:
SOAP contains some additional metadata for each dataset that we need to read in. 

- SOAP files are first generated and then compressed by a separate script. This means we save the type of compression to apply to the dataset, and whether the compression has been applied.
- We want all the arrays in SOAP to be the same length (the number of subhalos). However, we don't calculate all quantities for all objects, e.g. we have some quantities that are only calculated for subhalos with 100 bound gas particles. This means the arrays can have a large number of zeros. I'm now reading in the mask applied and adding that information to the array description.

kyleaoman:
Hi Rob, little thing that I noticed trying to experiment a bit with this, if I install normally (`pip install swiftsimio/`) I get:
```
>>> import swiftsimio as sw
  File "<stdin>", line 1, in <module>
  File "/cosma/home/durham/dc-oman1/.virtualenv/ssio/lib/python3.12/site-packages/swiftsimio/__init__.py", line 1, in <module>
    from .reader import *
  File "/cosma/home/durham/dc-oman1/.virtualenv/ssio/lib/python3.12/site-packages/swiftsimio/reader.py", line 14, in <module>
    from swiftsimio import metadata
  File "/cosma/home/durham/dc-oman1/.virtualenv/ssio/lib/python3.12/site-packages/swiftsimio/metadata/__init__.py", line 9, in <module>
    from .soap import soap_types
ModuleNotFoundError: No module named 'swiftsimio.metadata.soap'
```
If I install via link (`pip install -e swiftsimio/`), I don't have this problem. Should just be a matter of including in the pyproject.toml file, I think.

JBorrow:
Gotta add the submodule to the list in the pyproject.toml

kyleaoman:
When you get to the stage of looking at the masking, could I make a request? I know that it was planned to be able to apply spatial masks like we do with particle data. Could we also have an option to read a user-defined slice of the subhalo table (might have to handle subhalo and fof tables separately? can't remember the structure in SOAP hdf5 files and the sample seems to have moved/vanished). It's possible that there's nothing extra to be done since with a `spatial_only=False` mask you can do virtually anything, just want to flag it as something to check. My use case is that I want to be able to specify that I'm interested in subhalo 12345 (using a mask makes sense for this), and then when I do:
```
data = sw.load(...)
data.bound_subhalo.enclose_radius  # or whatever other field
```
I would like to read only the single row of interest. I said slice above because there could be use cases (hypothetically) to read a range or other slice, and it's the same amount of work to implement this as masking to a single row. Like I said I don't think much is needed because the masking tools just about support this already, but I guess the code needs to understand what's an array of subhalo data vs fof data vs anything else.

robjmcgibbon:
> When you get to the stage of looking at the masking, could I make a request? I know that it was planned to be able to apply spatial masks like we do with particle data. Could we also have an option to read a user-defined slice of the subhalo table (might have to handle subhalo and fof tables separately? can't remember the structure in SOAP hdf5 files and the sample seems to have moved/vanished). It's possible that there's nothing extra to be done since with a `spatial_only=False` mask you can do virtually anything, just want to flag it as something to check. My use case is that I want to be able to specify that I'm interested in subhalo 12345 (using a mask makes sense for this), and then when I do:
> 
> ```
> data = sw.load(...)
> data.bound_subhalo.enclose_radius  # or whatever other field
> ```
> 
> I would like to read only the single row of interest. I said slice above because there could be use cases (hypothetically) to read a range or other slice, and it's the same amount of work to implement this as masking to a single row. Like I said I don't think much is needed because the masking tools just about support this already, but I guess the code needs to understand what's an array of subhalo data vs fof data vs anything else.

Sorry about moving the example catalogue, they're now at `/cosma8/data/dp004/dc-mcgi1/FLAMINGO/Runs/L1000N1800/DMO_FIDUCIAL/SOAP/HBTplus`

I can make a new method for SWIFTMask called something like `constrain_index`. I don't think it should matter if it's a FOF/SOAP/snapshot file, they should all work the same way.

kyleaoman:
Makes sense. In EAGLE we had `/FOF/...` datasets and `/Subhalo/...` datasets coexisting in the same file, couldn't remember offhand if this was the case in SOAP, which would require a bit more logic, but looking now, it's not, so it should indeed be a minimal addition. Thanks!

robjmcgibbon:
@JBorrow update on the current status of this in case you have time to look it over before the meeting next week

This change adds an attribute to the SWIFTMetadata class called "filetype" which can be either "snapshot", "FOF", or "SOAP". I haven't changed much when it comes to reading in the files, everything has just been made more generic so instead of requiring the data to be in groups named "PartTypeX", the groups can have any names. Masking appears to be working for the "SOAP" filetype, it's not supported for "FOF" since SWIFT does not output spatially sorted catalogues.

SWIFT has added two new attributes to it's units metadata - one to indicate if a value is stored as physical, another to indicate if a value can be transformed between comoving and physical. This was motivated by quantities such as BirthDensities, which it doesn't make sense to convert. These attributes are now read in when reading a file, and the cosmo_array has a `valid_transform` attribute, which throws an error if someones tries to convert something they shouldn't.

Things to be done:
- Add tests for the "FOF" and "SOAP" filetypes
- Add documentation for the "FOF" and "SOAP" filetypes
- Implement a writer for "FOF" and "SOAP" (not sure this is needed, other than being helpful for tests)
- Decide what to do with swiftsimio functionality that only make sense for snapshots (e.g. visualisation)

JBorrow:
I am not sure that there's any functionality that only makes sense for snapshots. Visualisation is still useful with halo catalogues; for instance I use that in a hacky way right now to create cross spectra with haloes...

swiftsimio/metadata/particle/particle_types.py
JBorrow: Are we not able to remove these for modern snapshots? As they contain the particle type names and fields named appropriately..? I'd like to use those so we don't have to add/handle additional or renamed particle types manually.

swiftsimio/reader.py
```diff
@@ -578,37 +578,46 @@ def postprocess_header(self):
...
-        Loads the particle types and metadata into objects:
+        Determine what type of file we are reading
+        """
+        if self.output_type == "FOF":
+            self.filetype = "FOF"
+        elif self.output_type == "SOAP":
+            self.filetype = "SOAP"
+        else:
+            # Default to snapshot
+            self.filetype = "snapshot"
```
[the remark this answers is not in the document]
JBorrow: I'd also like to recommend something like...

self.filetype = "snapshot"

if self.output_type in ["FOF", "SOAP"]:
  self.filetype = self.output_type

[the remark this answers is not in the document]
JBorrow: But in general this needs more explanation

swiftsimio/reader.py
```diff
@@ -977,13 +1023,21 @@ def load_field_compressions(self):
...
+            except KeyError:
+                is_compressed = True
             try:
                 comp = dataset.attrs["Lossy compression filter"].decode("utf-8")
+            except AttributeError:
+                # Compression is saved as str not bytes
+                comp = dataset.attrs["Lossy compression filter"]
             except KeyError:
                 # Can't load compression string!
                 comp = "No compression info available"
```
JBorrow, line 1038: Same here

swiftsimio/reader.py
```diff
@@ -1391,44 +1497,55 @@ def generate_dataset(particle_metadata: SWIFTParticleTypeMetadata, mask):
...
+    group = group_metadata.group
+    group_name = group_metadata.group_name
+
+    # Set nice name for group
+    if group_metadata.metadata.filetype == "snapshot":
+        group_nice_name = metadata.particle_types.particle_name_class[group]
+    elif group_metadata.metadata.filetype == "FOF":
+        group_nice_name = "FOFGroups"
+    elif group_metadata.metadata.filetype == "SOAP":
+        group_nice_name = metadata.soap_types.get_soap_name_nice(group)
```
JBorrow, line 1521: Why if we wrote all this code are we doing this three different ways for three catalogues that we wrote..?

PR #193, opened 2024-07-17
Handle another case where we might get a string or bytes.

kyleaoman:
Seems like there are a few places where guarding against getting a string when we expected bytes is already needed, just add one more that was preventing opening new soap catalogues on python3.12.4 on cosma.


kyleaoman:
Turns out this was fixed independently elsewhere and is now redundant.

PR #199, opened 2024-09-19
Add homogeneous attribute

robjmcgibbon:
For files where all arrays have the same length and order (SOAP and FOF) we might want to extract a single row (all the properties for a single object). This doesn't make sense for snapshots.


kyleaoman:
@robjmcgibbon straightforward, I could tick the review box on this if you want.

JBorrow:
I'm closing this as I am patching these changes into #201

JBorrow:
fwiw i added tests for constrain index and indicies

PR #209, opened 2024-10-15
changed units of current from statA to A

nickishch:
Changed current units (because for MHD we use A current units, not statA)

### DIFF
--- a/swiftsimio/metadata/unit/unit_types.py
+++ b/swiftsimio/metadata/unit/unit_types.py
@@ -9,7 +9,7 @@ unit_names_to_unyt = {
     "Unit mass in cgs (U_M)": unyt.g,
     "Unit length in cgs (U_L)": unyt.cm,
     "Unit time in cgs (U_t)": unyt.s,
-    "Unit current in cgs (U_I)": unyt.statA,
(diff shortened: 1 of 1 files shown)


nickishch:
> And this resolves all your problems?

yes, at least it agrees with what snapshots show us
1) in snapshots the conversion factor to CGS ( units of g/(A*s**2) ) is 2e-6 for the runs I had (as on the screenshot)
2) opening the same snapshot in swiftsimio with the fix gives magnetic fields in units 1e-6*Msun/(A*Gyr**2), and this converts to 2e-6*g/(A*s**2), this agrees with the snapshots

Though there is some Issue when converting  g/(A*s**2) to Gauss. unyt says that for some reason 1 gauss = 1 cm^0.5⋅g^0.5⋅s^−1 but on wikipedia we also can find that 1 gauss is 0.1*g/(A*s**2) or 10^-4*kg/(A*s**2) as they say
https://en.wikipedia.org/wiki/Gauss_(unit). 
In practice multiplying the result in g/(A*s**2) units by 10 gives MF in gauss, so I don't know if we need to devote a lot of time to this issue, how do you think?

Screenshot 2024-10-11 at 13 35 42

JBorrow:
Thank you!

PR #211, opened 2024-11-18
Add fields to __repr__

robjmcgibbon:
Print out the available fields when examining an object in interactive mode


kyleaoman (approved):
Have been meaning to do this for ages, thanks Rob. Just added in some examples in the docs.

PR #212, opened 2024-11-22
Add support for np.cbrt on cosmo arrays

yuyttenhove:
Now that unyt [supports `np.cbrt`](https://github.com/yt-project/unyt/pull/491), since version [3.0.2](https://github.com/yt-project/unyt/releases/tag/v3.0.2), we can add support in swiftsimio as well.

We can then also get rid of the hack I used in the nearest neighbours slicing backend (which b.t.w. contained a bug in for the case where no `volumes` dataset is present in the gas dataset).


yuyttenhove:
I also had to edit an unrelated conversion test that was using a protected attribute of the astropy cosmology object that appears to have been removed in the 7.0.0 release...

kyleaoman (approved):
Thanks for spotting that unyt added support for this and for the PR :)

PR #214, opened 2024-11-30
Unpack the boxsize as a cosmo_array.

### LINKED ISSUE #128: boxsize in metadata should be cosmo_array, not unyt_array?
In `reader.py` the `SWIFTMetadata` object unpacks the `BoxSize` from the header as:
https://github.com/SWIFTSIM/swiftsimio/blob/a157029803f5584585b123421440e85b5e72269c/swiftsimio/reader.py#L371-L373

It would make more sense for this to be a `cosmo_array` (in particular specifying the correct `comoving` flag). I'm in principle happy to make a PR, but after staring at the code for ~15 minutes I can't work out where to put this in. Could be in `reader.py` if this is just to be a one-time exception for the box size, like for the scale factor:
https://github.com/SWIFTSIM/swiftsimio/blob/a157029803f5584585b123421440e85b5e72269c/swiftsimio/reader.py#L534-L539
Or if there are other header fields that similar logic could apply to it would be better placed in `metadata/metadata/metadata_fields.py`?

kyleaoman:
Closes #128 

Finally got around to implementing this. It's set up to mimic the way units are applied to header fields.

One thing to consider is whether we're happy with how the non-cosmological case is handled. If the simulation is non-cosmological, then the `scale_factor` will default to 1.0 (because swiftsimio enforces that it exists). The `box_size` is then created as a `cosmo_array` with `comoving=True`, but that's acceptable because `scale_factor=1` so it will just silently convert to physical if it ever needs to.

If we're not happy with that, then some further introspection will be needed to figure out if the snapshot is non-cosmological.

### DIFF
--- a/swiftsimio/metadata/metadata/metadata_fields.py
+++ b/swiftsimio/metadata/metadata/metadata_fields.py
@@ -77,7 +77,7 @@ def generate_cosmo_args_header_unpack_arrays(scale_factor) -> dict:
     # should not be cosmo_array'd).
     cosmo_args = {
         "boxsize": dict(
-            cosmo_factor=cosmo_factor(a**1, scale_factor=scale_factor),
+            cosmo_factor=cosmo_factor(a ** 1, scale_factor=scale_factor),
             comoving=True,  # if it's not, then a=1 and it doesn't matter
             valid_transform=True,
         )
--- a/swiftsimio/metadata/objects.py
+++ b/swiftsimio/metadata/objects.py
@@ -113,14 +113,12 @@ class SWIFTMetadata(ABC):
         # items including the scale factor.
         # These must be unpacked as they are stored as length-1 arrays
 
-        header_unpack_float_units = (
-            metadata.metadata_fields.generate_units_header_unpack_single_float(
-                m=self.units.mass,
-                l=self.units.length,
-                t=self.units.time,
-                I=self.units.current,
-                T=self.units.temperature,
-            )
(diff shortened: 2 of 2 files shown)


tests/test_soap.py
```diff
@@ -60,7 +60,9 @@ def test_soap_can_mask_spatial_and_non_spatial_actually_use(filename):
     masses2 = data2.spherical_overdensity_200_mean.total_mass
 
     # Manually mask
-    custom_mask = (masses2 >= lower) & (masses2 <= upper)
+    custom_mask = (
+        unyt.unyt_array(masses2.to_value(masses2.units), masses2.units) >= lower
```
robjmcgibbon, line 64: Why is this needed, does masses2 not already support comparison?

robjmcgibbon (approved):
Looks good to me, apart from a minor question about the test change

tests/test_soap.py
kyleaoman: `lower` and `upper` are `unyt_quantity`s, not `cosmo_array`s. This is correct given the type hints of `constrain_mask` used further up in the test, but correctly triggers a warning on comparison with `masses2`. Perhaps `constrain_mask` should be modified to expect `cosmo_array`s. It would still work in that case but would raise warnings on receiving `unyt_quantity`s (which is what should happen in that case, I guess). If that happens then the change to `custom_mask` should be reverted (it will emit a warning in that case so shouldn't go unnoticed).

PR #215, opened 2024-12-05
Handle no-cosmo_factor case in cosmo_array ufuncs

### LINKED ISSUE #203: Bug in objects.py
In function `_return_without_cosmo_factor` this:

```
        elif (cf is not None) and (cf2 is not None) and (cf == cf2):
            # both have cosmo_factor, and they match:
            pass
```

should be:

```
        elif (cf is not None) and (cf2 is not None) and ((cf == cf2) or (cf is None and cf2 is None)):
            # both have cosmo_factor, and they match, or neither has cosmo_factor:
            pass
```

kyleaoman:
Previously the case where 2 `comso_array`s both had no `cosmo_factor` was not handled in `cosmo_array` class when checking argument compatibility for ufuncs. This PR remedies this.

Note that there was a mistake in the bug report, the parentheses were misplaced (but the comment in the bug report code specified the correct intent).

Closes #203


JBorrow:
Can we remove the pngs?

kyleaoman:
Entirely, or just as a cleanup step in testing? FWIW I don't think that those have anything to do with this PR, but can throw it in anyway if you want...

JBorrow:
I mean don't commit them to the repo as is done here... They are produced as part of the tests

kyleaoman:
Whoops. Fixed.

JBorrow:
I'll squash merge this then! Thanks.

PR #217, opened 2024-12-06
Comoving parameter of cosmo_array defaults to None instead of True

### LINKED ISSUE #168: default comoving attribute of cosmo_array should be None?
If we initialize a `cosmo_array` like this:

```python
cosmo_array(np.array([1]) * u.Mpc)
```

the `comoving` attribute defaults to `True`. If we try to convert to physical units we get an exception because it tries to reference attributes of the `cosmo_factor` which is `None`, but trying to convert to comoving works fine. This seems dangerous, wouldn't it be better to default to `None` when the situation is ambiguous? Alternatively returning a `unyt_array` instead wouldn't be unreasonble, but is probably more confusing to someone expecting a `comso_array`. Could do it anyway and produce a warning.

Implementing this would need some checking that nothing in swiftsimio relies on the default. Test coverage might be good enough for issues to raise their head if it is relied on, but then again they might not ;)

kyleaoman:
Title pretty much says it - it's a one-line code change. Rationale explained in related issue: closes #168 

Doesn't seem to break anything in the test suite, and this is a sufficiently fundamental change that I would have expected it to break something in there if it was going to. The main reader functions already set the parameter explicitly.

### DIFF
--- a/swiftsimio/objects.py
+++ b/swiftsimio/objects.py
@@ -728,7 +728,7 @@ class cosmo_array(unyt_array):
         input_units=None,
         name=None,
         cosmo_factor=None,
-        comoving=True,
+        comoving=None,
         valid_transform=True,
         compression=None,
     ):


robjmcgibbon (approved):
Looks good to me, I think it makes sense to set it to None now that we can throw the InvalidConversionError

PR #221, opened 2025-03-22
Handle previously unhandleable unyt operations

kyleaoman:
Previously operations like `u.Mpc * cosmo_quantity(...)` and `(1 * u.Mpc) * cosmo_quantity(...)` silently demoted to `unyt` return types because these would be handled by the parent class `__mul__` methods before we had any chance to intervene with our own `__rmul__`. 

I've opened https://github.com/yt-project/unyt/pull/572 that enables allowing a subclass (like our `cosmo_array` and `cosmo_quantity`) to add its own handling in cases like this. The current draft of that PR gives everything we need and at the time of writing this PR is ready to merge once unyt merges (and releases...) that. It's possible that the upstream PR will need some amendments, though, so will need to check that everything still works on our end. Currently the test suite (including some new tests) on this branch passes with the `allow_subclass_rmultiply` branch on my fork of `unyt`. I would expect the tests on this branch (`support_rmul`) to raise any issues when the time comes.


swiftsimio/objects.py
```diff
@@ -1656,6 +1657,99 @@ def from_pint(
...
+                    r.view(cosmo_quantity)
+                    if r.shape == ()
+                    else (
+                        r.view(cosmo_array)
+                        if isinstance(r, unyt_array) and not isinstance(r, cosmo_array)
+                        else r
+                    )
+                )
+                for r in result
+            )
```
JBorrow: Does this not just immediately get overwritten by the loop below?

PR #222, opened 2025-04-06
Add python3.13 to CI testing and update release workflow

### LINKED ISSUE #197: Deploy flow broken
Probably just needs a python revision

kyleaoman:
Trying to switch on python3.13 in testing, to see what (if anything) needs attention to support it. Looks like nothing breaks.

Also modernizing the test and deployment workflows a bit so that we can use the deployment workflow again. Closes #197


kyleaoman:
@JBorrow looks like nothing breaks...

kyleaoman:
@JBorrow I've configured publishing on the PyPI side. This repo needs an environment added called `pypi` for it to work, though, and I don't have permissions for that. Shouldn't need any special configuration of the environment, it just needs to exist. Here's what it looks like for `swiftgalaxy`, for instance:
Screenshot from 2025-04-16 14-45-40

PR #230, opened 2025-04-18
Use cell bounding box metadata

### LINKED ISSUE #88: Test and document wrapped masking
If we use a spatial only mask, the box is wrapped correctly (it seems). This is undocumented behaviour, and is untested. It would be good to fix both of thsoe.

### LINKED ISSUE #121: Add support for partilcle bounding boxes in the snapshot meta-data
This SWIFT PR https://gitlab.cosma.dur.ac.uk/swift/swiftsim/-/merge_requests/1525 introduces particle bounding boxes in each top-level cell's meta-data. This can be useful to speed-up the reading of data if one know that it is unnecessary to read neighbouring cells.

Is the format proposed in the SWIFT PR sensible? Can swiftsimio exploit it or should I add something else to ease this?

### LINKED ISSUE #177: Test data is very outdated
Our test data is extremely outdated. This means the test suite passes when some of our features don't work with new data (e.g. subset snapshots).

We need to re-run the small cosmo volume and update the data on cosma dataweb.

kyleaoman:
- [x] Implemented using the cell bounding box metadata (min particle position and max particle position), if present. If not behaviour changes: we pad the selection by 0.1 cell length to make sure that we get any particles that have drifted out of their cells. If they've drifted by >0.1 cell length... too bad.
- [x] Added some tests as suggested by #88. The old test data doesn't include the new metadata (min/max positions). Which leads to...
- [x] Made some new test data, addressing #177. It's located at `/cosma5/data/Swift/web-storage/IOExamples/ssio_ci_04_2025`. I've left the old test data in place in case that's used by anyone else. There's a readme in the new test data directory with details. Next time we decide we need new test data we can just make a new subdirectory. I've also included some of the old test data files (symlinked in) as "legacy" examples to make sure we don't break backwards compatibility. I've also made sure that we have both a "single" and "distributed" file
- [x]  Set up fixtures to either test with both the old and the new single and distributed test files (so we can ensure backwards-compatibility and compatibility with both types), or just the new single-snapshot file where this is not relevant.
- [x] Abandoned the `cosmo_volume_example.hdf5` test data file and use the new single-file snapshot instead.
- [x] Documented the masking behaviour with respect to periodic boxes as suggested in #88.
- [x] Made the cell padding length configurable (including able to be disabled) in cases where the cell bounding box metadata is absent. 

Closes #121 
Closes #88 
Closes #177


JBorrow:
I know it's annoying but it would be helpful in the future to split things up into separate PRs (e.g. the changes to the test data).

I'm a bit confused here - are we now going to load 9 cells every time we want to load just one? This could be quite significant in e.g. FLAMINGO setups.

kyleaoman:
Yes, in hindsight I would have split this up. I didn't plan on updating the test data for this, but eventually realised that the existing examples didn't have the `MinPosition` and `MaxPosition` metadata. Then of course updating the sample data didn't "only" add those metadata fields, but revealed some other bugs that had crept in due to the stale test set. I think that this is the last rat's nest to untangle that I can see in the existing github issues - at least my intention is to keep things more compartmentalised in the future. I don't reeeeally want to, but I could go back and split this if it's needed.

As to how much gets read - if it's an old snapshot, then yes, the minimum read is now 9 cells. That's the only way to guarantee that you get all of the particles spatially located in the cell.

For newer snapshots then if you ask for a region that is exactly the size of a cell, it depends on what the metadata says. If no particles from neighbouring cells have drifted into the target cell, you'll only read one. If particles from the cell above and the one to the left have drifted in then you'll read 3 (not 9!). This means that if you're interested in an object within the cell, you should define the region to be the box around the object that you want, and not "snap" to the cell, because there's enough information to check that even though particles may have drifted into the target *cell*, we might still be sure that they haven't drifted into the target *region* within that cell.

The above is all outlined in the updated narrative docs.

I don't know whether the flamingo snapshots had `MinPosition` and `MaxPosition` included. Picking a couple of the big runs at random it looks like they did not. Of course reading 1 cell is the worst-case scenario of a 9:1 increase. If you read say a 20x20x20 cell chunk then it's 20^3:22^3 ~30% more I/O (instead of 900%).

If you think that there's a use case for risking missing out some particles in the region that you think that you loaded then I could add a flag, something like `mask(region, spatial_only=True, pad=True)`. I think the padding should be on by default, though.

JBorrow:
I guess I'm wondering about the case where we have a region selection that's much smaller than an individual cell (that's kinda common with flamingo). This then goes back to your point about the region selection. Imagine you are reading a small volume right in the center of the cell. The probability that there has been drift into there is tiny... It just seems that we're giving up a lot of efficiency for perhaps little gain, though I obviously agree your implementation is the safe and correct thing to do. For analysis, you may want to do that, but for basic visualization you are potentially increasing your runtime by a factor of 9.

I am in favour of adding this as an optional parameter that defaults to `True`.

This actually sounds like an upstream bug that needs to be fixed at some point. I'm still in favour of writing out particles ordered as an oct-tree :).

MatthieuSchaller:
I would worry about having to read 27x data that expected being the default behaviour. 

My suggestion would be that it should be the responsibility of downstream tools to choose the cells they need wisely. If I get an object no where near a cell boundary via SWIFTgalaxy (say) then that package could decide to not load more than the cell of interest. Or load more if the object is close to the edge. 
With the bounding box of each object (soon in FOF) and the max radius in SOAP one can easily do that. 

I think `swiftsimio` remaining "dumb" is best.

JBorrow:
Sorry - yes - 3D space is indeed three dimensional. It's a factor of 27!

kyleaoman:
@MatthieuSchaller it's not so obvious to me what the right default behaviour is here. If we don't pad (or don't pad conservatively enough) then the risk is some very hard to debug problems for users. As a simple example, let's say I want to measure the variance in the matter density in a top-hat filter of some size. I don't want to load the whole box, so I load a region at a time of the size of my filter. I can easily realise that there are some extra particles outside of the region that I asked for (because my region didn't align perfectly with cells), so I discard particles outside the region after reading them in, and calculate the density. Without the padding, though, I could be missing some particles that have drifted into the region, and make an incorrect measurement. It's very hard to notice this (would have to read a larger region/whole box and compare). Counting on users reading the documentation closely enough to get the information from there without an obvious hint is probably too optimistic.

This is a problem that will gradually go away over time since all newer snapshots will have the `MinPosition` and `MaxPosition` annotations and therefore the code reads in exactly what is requested. The question is therefore what default behaviour to adopt for older snapshots while they're still in use. I've assumed 1 cell length as padding, but that's not technically perfectly safe since in theory a particle could drift arbitrarily far from its cell centre by the time the snapshot is written, and "always read the whole box" is not a solution. Do you have any diagnostics for how many and how far particles drift out of their cells? I assume that this is basically impossible to quantify since the cell count, size, particles per cell, snapshot cadence, etc etc are all configurable and problem-dependent.

Perhaps we'd be happier with a smaller pad as default - say 0.1x or 0.2x the cell length This way if you pick a region in the centre of a cell you only read that cell, but if it's close enough (within 0.1 or 0.2) to a neighbouring cell you also read the neighbour. If you try to read a region exactly the size of a cell you'll still read 27x neighbours, but I would argue that this is desirable.

Related, do we want to always emit a warning when the `MinPosition` and `MaxPosition` metadata are absent and `constrain_spatial` is used, since we can never fully guarantee that the operation is "safe"?

kyleaoman:
There is now a mechanism for users to adjust the padding (ignored for newer snapshots where the cell bounding box is known), and tests for that. Once we agree on a desired default behaviour it's trivial to implement.

MatthieuSchaller:
What simulations are affected? Is it mainly a legacy problem or are colibre runs also in a state where they need 27x more data?

kyleaoman:
I checked a few Colibre runs, they seem to have the metadata present. I also checked a few Flamingo runs, there the metadata is absent.

kyleaoman:
@MatthieuSchaller @JBorrow @robjmcgibbon to make this concrete, I propose that:

1. If the cell bounding box metadata is present then that is used to determine what cells to read. Any bbox that touches the region gets the cell selected, period. 
2. If the cell bbox metadata is absent then a warning is emitted and a default pad of 0.1 cell side length is applied. The warning mentions how to configure the pad to be very safe (pad by 1 cell side length) or minimize i/o (no pad).

In case it wasn't clear, padding by <1 cell helps because the region is generically not aligned with the cell grid, and the pad is not adding adjacent cells to be read but just making each cell slightly bigger when deciding whether its bbox touches the region.

I feel like this strikes a reasonable balance between accuracy and minimizing unnecessary overheads. Does anyone want to voice an objection to the above? If not I'll wrap up the modifications and call it (actually) ready for review.

JBorrow:
The only tricky thing here I think would be to make sure that the extra padding doesn't lead to extra padding on things like the particle-level mask you can ask for. As long as the padding is only used for cell selection I see no issue with adding a small pad to the region to account for cases like this.

kyleaoman:
It's implemented directly in the [spatial masking function](https://github.com/SWIFTSIM/swiftsimio/blob/bd6bcd1f0bc70fa0e4309c61ca5c641641271e5d/swiftsimio/masks.py#L421), so it can't "leak out" into other modes. Reading catalogue files (soap, fof) where particles drifting does not happen is also guarded (never padded).

MatthieuSchaller:
Ok. I agree with this last proposal. Better be safe for the old data. The new sims with more metadata will only read what is required. 

What about the buffer size in the old case? Are we not always reading one full extra cell? I  must be missing something about the 0.1

kyleaoman:
@MatthieuSchaller Hmm I tried in words before, let's go with the diagram:



The pad is applied cell by cell (so that when the metadata is available it can be variable). If the user asks for a region exactly aligns with the cell grid then both padding options will read the same set of cells. In this diagram the region is not along grid vertices so we save on I/O by padding less.

MatthieuSchaller:
Thanks. I am with you now. That sounds good to me.

JBorrow (approved):

PR #231, opened 2025-04-21
Fix variable re-use in power spectrum

### LINKED ISSUE #218: Computation power spectra with folding
Hi,

The way the power spectra with folding are computed currently is by changing the positions of particles according to
```
box_over_fold_x = box_x / fold
inv_box_over_fold_x = 1.0 / box_over_fold_x
x_pos = (x_pos % box_over_fold_x) * inv_box_over_fold_x
```
with `fold = 2.0^folding`, i.e. (from swiftsimio readthedocs)
image

Should this not be 
`x_pos = (x_pos % box_over_fold_x) * fold` ?

For example, without folding (folding = 0 and fold =1) the positions are now converted to 
`x_pos = (x_pos % (Lbox/1) ) * 1 / Lbox = x_pos / Lbox`

JBorrow:
Fixes #218

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ htmlcov/*
 
 # Formatting venv
 black_formatting_env/*
+
+# Leftovers
+*.png
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ black_formatting_env/*
 
 # Leftovers
 *.png
+*.hdf5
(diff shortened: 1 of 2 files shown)


.gitignore
kyleaoman, line 1: Want to add `*. hdf5` here too?

swiftsimio/visualisation/power_spectrum.py
kyleaoman, line 1: Is "same units" required? Or is "compatible units" enough? `cosmo_array` should be clever enough now for the latter. Might want to check that mixed physical & comoving inputs give intuitive outputs, do you want to explicitly coerce the box size to match the comoving flag of the coordinates?

kyleaoman (approved):

PR #242, opened 2025-06-09
Fix a deprecated import and an unclosed file in tests

kyleaoman:
Just two straightforward fixes for warnings cropping up in the test suite.


JBorrow:
It would be wise to pin a known version of astropy where this import works; it was the old style because that was the only way to import the `Cosmology` object.

PR #248, opened 2025-07-17
Fix critical bug in power spectra

JBorrow:


kyleaoman:
@Darkmattermaster070 just so you know, this is being tracked in #253

Darkmattermaster070:
@kyleaoman Thank you!

PR #273, opened 2025-10-28
Fix off-by-one bugs in visualisation

kyleaoman:
The 0th row and 0th column of images (and 0th element along the 3rd dimension for volume renders) was getting extra contributions from periodic copies of particles when `periodic=True` in all visualisation functions. The logic incorrectly assumed that e.g. `int(-0.5) == -1` (instead of truncating to `0`).

- [x] Fixed the initial off-by-one bug in the particle skipping logic across all visualisation functions, plus a bonus one in the `"histogram"` backend.
- [x] Fixed another similar bug in `"subsample"` and `"subsample_extreme"`.
- [x] Still need to add some test coverage.
- [x] There's still a bug affecting the corner near (0,0) for the `subsampled` and `subsampled_extreme` backends, visible in residuals of the `big_img` minus a tiled `ref_img` in `test_visualisation.py`'s `TestProjection: test_equivalent_regions`. Probably another off-by-one type bug...

### DIFF
--- a/swiftsimio/visualisation/projection_backends/subsampled.py
+++ b/swiftsimio/visualisation/projection_backends/subsampled.py
@@ -190,6 +190,11 @@ def scatter(
 
                     if not (
                         overlaps_left or overlaps_right or overlaps_down or overlaps_up
+                    ) and (
+                        particle_cell_x >= 0
+                        and particle_cell_x <= maximal_array_index
+                        and particle_cell_y >= 0
+                        and particle_cell_y <= maximal_array_index
                     ):
                         # Very simple case - no overlaps.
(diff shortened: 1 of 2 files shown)


kyleaoman:
I tightened up the tolerance a lot on some of the existing tests, probably just differences in rounding on the CI machine giving residuals of 1e-4 instead of 1e-5 or something like that. Will figure it out.

kyleaoman:
That's pretty weird. It looks like something in serialization in the parallel mode, I guess that could be platform dependent somehow. Presumably can be fixed... If you can have a look at it in the meantime anyway that'd be great so we can hopefully get the fix out soon.

Issue #277, opened 2025-10-31
Add a test for #276


kyleaoman:
This was rushed in to fix a crash, should add test coverage - this issue is a reminder.

robjmcgibbon:
Would you be alright if we add a colibre subset snapshot to the test data? None of the test data currently has `SubgridScheme/GrainToElementMapping`

PR #280, opened 2025-11-07
Add a colibre test dataset

robjmcgibbon:
Addresses https://github.com/SWIFTSIM/swiftsimio/issues/277

There's a colibre test file at `/cosma8/data/dp004/dc-mcgi1/COLIBRE_L006m7.hdf5`, which would need to be copied to the webstorage at `/cosma5/data/Swift/web-storage/IOExamples` (I don't have permission). This is just a copy of the L006m7 snapshot.

I had to make some minor updates to existing tests.


kyleaoman:
@robjmcgibbon do you understand the test failures? Seems weird that it passes on 3.11 but fails on 3.10 based on a key missing in the cosmology? Maybe old versions of astropy don't have `Omega_ur`?

PR #281, opened 2025-11-18
Add support for remote snapshots via hdfstream

jchelly:
This is a continuation of #252. There are enough changes to both swiftsimio and this branch that I think it makes sense to start again from the current master.

Kyle's changes in #258 have enabled (what I hope is) a cleaner way to approach this. I've modified swiftsimio so that the user can pass in a file handle instead of a filename. In this case swiftsimio should never open or close any files itself. The user takes responsibility for closing the file. We can then pass in a hdfstream.RemoteFile instead of a h5py.File to read remote snapshots and swiftsimio doesn't need to handle the server URL or authentication credentials.

This is still work in progress (e.g. there are a couple of test failures).

### DIFF
--- a/.github/actions/start-hdfstream/action.yml
+++ b/.github/actions/start-hdfstream/action.yml
@@ -12,7 +12,7 @@ inputs:
   image:
     description: 'Server container image to use'
     required: true
-    default: ghcr.io/jchelly/hdfstream-api:0.0.5
+    default: ghcr.io/jchelly/hdfstream-api:0.0.6
 runs:
   using: "composite"
   steps:
--- a/.github/workflows/lint_and_test.yml
+++ b/.github/workflows/lint_and_test.yml
@@ -55,7 +55,7 @@ jobs:
       with:
         data-dir: ./test_data
         virtual-prefix: test_data
-        image: ghcr.io/jchelly/hdfstream-api:0.0.5
+        image: ghcr.io/jchelly/hdfstream-api:0.0.6
     - name: Test with pytest
       run: |
         pytest --hdfstream-server=http://localhost:8080/hdfstream


jchelly:
The tests now pass when run via github but I've doubled the runtime by running a lot of the tests three times. I might need to be a bit more selective. Oddly, the subset writer test is significantly faster on remote files: 25s vs 70s running the EagleDistributed.hdf5 test on Cosma). When using a local file 90% of the time is spent in ~5000 read_direct() calls.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 astropy
 h5py
+hdfstream
```
kyleaoman: Same here.

kyleaoman: Also check whether it's a requirement to build the docs (`docs/requirements.txt`).

swiftsimio/accelerated.py
```diff
@@ -453,6 +418,154 @@ def read_ranges_from_file_chunked(
...
+        HDF5 dataset to slice data from.
+
+    ranges : np.ndarray
+        Array of ranges (see :func:`ranges_from_array`).
+
+    output_shape : Tuple
+        Resultant shape of output.
+
+    output_type : type, optional
+        ``numpy`` type of output elements. If not supplied, we assume ``np.float64``.
```
kyleaoman: Should be possible to xref
```
:mod:`numpy`
```

swiftsimio/accelerated.py
```diff
@@ -453,6 +418,154 @@ def read_ranges_from_file_chunked(
...
+            output_sorted[offset : offset + n, ...] = output[start:stop, ...]
+            offset += n
+        output = output_sorted
+
+    if not output.dtype.isnative:
+        # The data type we have read in is the opposite endian-ness to the
+        # machine we're on. Convert it here, to save pain down the line.
+        output = output.byteswap(inplace=True).newbyteorder()
+
+        if not output.dtype.isnative:
```
kyleaoman: It would be a bit unfortunate to download a lot of data, go through the trouble of byteswapping and only then discover that the data are useless. Can this check be moved up? Maybe read an initial minimal amount of data, check that it's going to work, and then proceed?

swiftsimio/file_utils.py
```diff
@@ -0,0 +1,104 @@
...
+from ._handle_provider import HandleProvider
+
+import h5py
+import hdfstream
+from pathlib import Path
+
+
+def is_soft_link(obj: h5py.Group | h5py.Dataset | h5py.SoftLink) -> bool:
+    """
+    Return True if obj is a soft link.
```
kyleaoman:
```
``True``, ``obj``
```
There are many code fragments that should be quoted in `file_utils.py`.

swiftsimio/reader.py
```diff
@@ -403,7 +406,14 @@ class __SWIFTNamedColumnDataset(object):
...
 
-    def __init__(self, field_path: str, named_columns: list[str], name: str) -> None:
+    def __init__(
+        self,
+        field_path: str,
+        named_columns: list[str],
+        name: str,
+        handle: h5py.File | None = None,
+    ) -> None:
+        super().__init__(handle.filename, handle=handle)
```
kyleaoman: Doesn't this crash if `handle` is `None`?

swiftsimio/subset_writer.py
```diff
@@ -156,7 +157,7 @@ def find_links(
     for key in keys:
         subpath = f"{path}/{key}"
         dataset = input_file.get(subpath, getlink=True)
-        if isinstance(dataset, h5py.SoftLink):
+        if is_soft_link(dataset):
```
kyleaoman: Part of making hdfstream an optional dependency could look like:
```
def is_soft_link(dataset):
    try:
        import hdfstream
    except ImportError:
        return isinstance(dataset, h5py.Dataset)
    else:
        return isinstance(dataset, (h5py.Dataset, hdfstream.RemoteDataset))  # or call such a function defined by hdfstream
```

swiftsimio/accelerated.py
jchelly: I think the danger here is if we get a compound type where the components aren't all the same endian, so the dataset is not native and byte swapping doesn't make it native. We can probably create a scalar of the same type as the dataset and run the check on that.

I wasn't sure what to do about endian-ness on the server. At the moment it always returns little endian data so we shouldn't get mixed types but we will need to byte swap on big endian clients.

jchelly: Yes, that type of thing. E.g.
```
x = np.ones(1, dtype=[("a", ">f4"), ("b", "<f4")])
print(x.dtype.isnative)
print(x.dtype.newbyteorder().isnative)
```
prints False for both cases. The server wont return data like this but h5py could if you went out of your way to set it up. Maybe we don't need to do anything, unless there's some more likely case that fails this check?

tests/conftest.py
```diff
@@ -57,13 +83,85 @@ def _requires(filename: str) -> str:
...
+
+
+access_methods = [
+    open_local_with_filename,
+    open_local_with_handle,
+    open_with_hdfstream,
+]
+
+
+filenames = [
```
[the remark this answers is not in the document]
jchelly: Ok, done. It didn't occur to me that I could do that for some reason.

swiftsimio/reader.py
jchelly: Yes, that's clearly not right. If __SWIFTNamedColumnDataset is going to be a HandleProvider then the filename needs to be passed in too. I can't quite recall why it needed to be a HandleProvider now. Will look into that.

jchelly: To go back to the original issue: we could fix it by just changing the docstring and annotations to say the handle must not be None. Each __SWIFTNamedColumnDataset instance should always have a parent SWIFTDataset to manage the handle. I assume the underscore name means that users shouldn't be instantiating a __SWIFTNamedColumnDataset directly.

kyleaoman: Yes that's right about the underscore - you can be quite stringent about how this is supposed to be initialised. Since it's only supposed to be used internally we have control.

tests/test_subset_write.py
```diff
@@ -72,8 +74,14 @@ def test_subset_writer(snapshot_or_soap):
     Writes a subset of the input file to a new file
     and compares result against masked load of the original file.
     """
+    # Get the name of the input test file
```
kyleaoman: `getattr(snapshot_or_soap, "filename", str(snapshot_or_soap) if isinstance(snapshot_or_soap, Path) else snapshot_or_soap)`, I guess, avoids the `str(str)` case...

swiftsimio/__init__.py
```diff
@@ -172,10 +171,8 @@ def load(filename: str | Path, mask: SWIFTMask | None = None) -> SWIFTDataset:
     SWIFTDataset
         Dataset object providing an interface to the data file.
     """
-    if isinstance(filename, str):
-        filename = Path(filename)
-
-    with h5py.File(filename, "r") as handle:
+    with open_path_or_handle(filename) as handle:
+        filename = Path(handle.filename)
```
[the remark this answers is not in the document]
jchelly: Done.

swiftsimio/accelerated.py
```diff
@@ -388,7 +368,7 @@ def read_ranges_from_file_chunked(
         Resultant shape of output.
 
     output_type : type, optional
-        :mod:`numpy` type of output elements. If not supplied, we assume ``np.float64``.
+        :mod:`numpy` type of output elements. If not supplied, we assume :py:attr:`numpy.float64`.
```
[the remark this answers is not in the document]
jchelly: Also done.

swiftsimio/_file_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    obj : str, Path or h5py.File
+        A path to a file or a file handle object.
+
+    Returns
+    -------
+    tuple[Path, h5py.File]
+        Tuple with the path and file handle.
+    """
+    if isinstance(obj, (str, Path)):
+        filename = Path(obj)
```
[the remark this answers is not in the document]
jchelly: That seems a bit opaque to me. The constructor for Path does accept other path objects, so maybe it's ok as is?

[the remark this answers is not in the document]
kyleaoman: It does end up being a bit awkward to handle. There's the very marginal advantage of avoiding
```
filename = Path("/")
split_filename, split_handle = split_path_or_handle(filename)
assert split_filename is filename  # AssertionError, might expect this to hold in this case
```
but probably not worth the mess in the code.

tests/conftest.py
```diff
@@ -2,19 +2,45 @@
...
 import subprocess
+import itertools
 import pytest
 from collections.abc import Generator
 import numpy as np
 import unyt
+import h5py
 from swiftsimio import Writer
 from swiftsimio.units import cosmo_units
+try:
```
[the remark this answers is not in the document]
jchelly: Is there an advantage to find_spec()? It seems like it would help if I wanted to check for a module without importing it, but here I'm immediately importing the module afterwards.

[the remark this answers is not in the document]
JBorrow: Yes I believe the rest of swiftsimio uses these `HAVE_X` booleans (see https://github.com/SWIFTSIM/swiftsimio/blob/master/swiftsimio/optional_packages.py). What we do there is offer no-op versions of the functionality.

I'd like to see that guarding pattern used here, i.e.:

```
if HAVE_HDFSTREAM:
  continue with my life
```
rather than
```
if hdfstream is not None:
```

jchelly:
I tried running the tests locally using the same docker image that runs on github and that hangs too. I think something might be going wrong with reading the body of the dataset so the response gets truncated, and the python client isn't handling that properly.

kyleaoman (approved):

kyleaoman:
The optional dependency makes sense, thanks John, this will hopefully be a very useful feature!

PR #291, opened 2026-01-29
Fix `cosmo_array.squeeze`

### LINKED ISSUE #238: `squeeze` on `cosmo_array` should return `cosmo_quantity`
Just as an arbitrary example encountered while working on something else:

```
>>> sg.halo_catalogue.input_halos_hbtplus.track_id
cosmo_array([162], dtype=uint64, units='(dimensionless)', comoving='False', cosmo_factor='a**0.0 at a=0.9999999999999997', valid_transform='False')
>>> sg.halo_catalogue.input_halos_hbtplus.track_id.squeeze()
cosmo_array(162, dtype=uint64, units='(dimensionless)', comoving='False', cosmo_factor='a**0.0 at a=0.9999999999999997', valid_transform='False')
```

Since the result is 0-dimensional we should get a `cosmo_quantity` but it's a `cosmo_array`.

kyleaoman:
Really simple fix for this one. Closes #238 

This is also an open issue in pure `unyt`: https://github.com/yt-project/unyt/issues/627 but no reason that we need to wait for that to be solved, this patch is fine in isolation.

### DIFF
--- a/tests/test_cosmo_array.py
+++ b/tests/test_cosmo_array.py
@@ -1114,6 +1114,8 @@ class TestCosmoQuantity:
         )
         assert ca.squeeze().ndim == 0
         assert isinstance(ca.squeeze(), cosmo_quantity)
+        assert np.squeeze(ca).ndim == 0
+        assert isinstance(np.squeeze(ca), cosmo_quantity)
 
 
 class TestCosmoArrayCopy:


kyleaoman:
Whoops, should probably make sure `np.squeeze` is ok too...

PR #292, opened 2026-02-05
Fix check for dropping to single cell in volume render

FHusko:
The volume render was sometimes missing particles altogether because their smoothing kernels did not encompass any cell centers. This is due to an incorrect one-half factor instead of sqrt(3) / 2 when checking how big the kernel is relative to the cell sizes. 

I did not modify the `scatter_limited_z` function, since that one isn't used at the moment and is not working anyways as far as I am aware.

### DIFF
--- a/swiftsimio/visualisation/volume_render_backends/scatter.py
+++ b/swiftsimio/visualisation/volume_render_backends/scatter.py
@@ -96,7 +96,7 @@ def scatter(
     float_res_64 = np.float64(res)
 
     # If the kernel width is smaller than this, we drop to just PIC method
-    drop_to_single_cell = pixel_width * 0.5
+    drop_to_single_cell = pixel_width * np.sqrt(3.0) / 2.0
 
     # Pre-calculate this constant for use with the above
     inverse_cell_volume = float_res * float_res * float_res
@@ -317,6 +317,7 @@ def scatter_limited_z(
     float_res_z_64 = np.float64(res_z)
 
     # If the kernel width is smaller than this, we drop to just PIC method
+    # TODO: Check that this actually works!
     drop_to_single_cell = pixel_width * 0.5
     drop_to_single_cell_z = pixel_width_z * 0.5
 


FHusko:
I think the check in `scatter_limited_z` would need to have the ratio of cell sizes in there somewhere (since the cell size along x and y is different from z). But perhaps we should worry about that once the function itself is working?

FHusko:
I'm not sure what's wrong with `scatter_limited_z`, @JBorrow will probably know.

I'll update here once I've done something on a regression test.

FHusko:
@kyleaoman Not yet, sorry! I also got ill recently which set me back a bit.

kyleaoman:
No problem, sorry to hear you haven't been well - no rush of course, might just send the occasional reminder.

kyleaoman:
@FHusko nudge?

FHusko:
@kyleaoman Sorry about the delays, I'm not sure if I'll have time to set up the unit test any time soon. I'm pretty sure things work, or at the very least work better with this fix (so do you need confirmation by a unit test?) because I've been running Partridge with this version for a while in different configurations and can confirm I don't get any particles that are skipped with this fix, whereas otherwise they can be.

kyleaoman:
@JBorrow do you know what's going on with the `scatter_limited_z` function? I read it carefully and if I follow (1) it's not actually callable through `volume_render` so it's effectively unusable and (2) it still assumes cubic cells (i.e. they are not long and skinny along the z-axis) but also doesn't really deal with the fact that you might want periodic boundaries in x and y but not z? Does it make sense to just remove the function? It's pretty similar to the basic `scatter` if it wants a proper re-implementation later...

Once that's settled I can wrap up this PR pretty easily if Filip doesn't have time.

JBorrow:
I think `scatter_limited_z` was something that I was testing for volume rendering but never made it into a full implementation. As far as I am concerned, it can be removed.

PR #293, opened 2026-02-05
Get metadata from user input instead of making assumptions in Writer.

### LINKED ISSUE #267: Remove obsolete hardcoded metadata definitions
We've been reading metadata from the SWIFT files for a long time, but we still have:
```
swiftsimio/metadata/units/unit_fields.py
swiftsimio/metadata/cosmology/cosmology_fields.py
```

Both of these are tagged as: `Deprecated v3.10.0`.

Can these be removed? Or are they needed to read (very) old snapshot files?

kyleaoman:
Refactored out all of the hardcoded unit metadata. The idea is that the `Writer` class now expects the user to provide `cosmo_array` data and get the metadata from there. The _dimensions_ of the required fields are still hard-coded (seems reasonable...), so those are still validated, and the data are stored in a consistent unit system (can still choose this by passing a `unit_system` to the `Writer`). Left to do:

- [x] Make sure there are clear prompts for what to do if a user passes `unyt_array` data as they would have been able to previously.
- [x] The tests of the `Writer` class were... not good. The existing ones are improved, and I've added coverage related to this PR (mostly checking behaviour on valid/invalid assignments to required fields). This has led to several new internal checks in the `Writer` class.
- [x] It seems like we don't allow writing non-required fields. For instance you can't write, say, gas metallicity using the `Writer`. This seems to be the intention (could be changed later), should make sure this is clearly documented.
- [x] Update narrative docs.
- [x] Make a review pass over docstrings. 
- [x] Refactor `units.py`.

Points for review:
 - There's some memory overhead, particularly around checking particle ID uniqueness (concatenate duplicates all IDs, then unique is expected to duplicate again). Are we ok with this? I couldn't see a way to test uniqueness without at least a single new array (e.g. could in-place sort as a starting point, but then need to store an index array to unsort at minimum).
 - I don't see any straightforward way to validate that a user is assigning something that will be written. E.g. they might reasonably expect that setting `w.gas.metallicity = cosmo_array(...)` to sensible values will make them appear in the output file, but these are silently ignored (only "required" fields are written). This is clearly documented - does this seem like enough?
 - Do we want to rename `snapshot_writer.py` to `ic_writer.py`?
 - The content of `swiftsimio.units` only seems to be used in the context of the `Writer`. I've relocated it to `swiftsimio.metadata.writer.unit_systems` - happy with this?

Closes #267

### DIFF
--- a/swiftsimio/snapshot_writer.py
+++ b/swiftsimio/snapshot_writer.py
@@ -672,10 +672,16 @@ class SWIFTSnapshotWriter(object):
         else:
             self.unit_system = unit_system
 
-        assert len(boxsize) == dimension, (
-            f"boxsize {boxsize} does not have length equal to number of dimensions "
-            f"({dimension})."
-        )
+        try:
+            if len(boxsize) != dimension:
+                raise ValueError(
+                    f"boxsize must have length equal to number of dimensions, {boxsize=},"
+                    f" {dimension=}."
+                )
+        except TypeError:  # len() of unsized object
+            raise ValueError(
+                "boxsize must be a cosmo_array of length equal to spatial dimensions."
+            )
         self.boxsize = np.copy(boxsize, subok=True)
         self.boxsize.convert_to_base(self.unit_system)
 
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -3,7 +3,7 @@
 import os
 import pytest
 import numpy as np
-from swiftsimio import load, cosmo_array, Writer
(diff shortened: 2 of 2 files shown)


kyleaoman:
Whoops found a little bug while updating swiftgalaxy for these changes. Bug is patched but should add a couple of tests for setting the boxsize and unit tests of `generate_smoothing_lengths`.

kyleaoman:
@JBorrow if you're happy with this is can be merged.

PR #302, opened 2026-03-18
Fix nearest_neighbour slice visualisation backend

### LINKED ISSUE #274: `nearest_neighbour` backend for slice images doesn't wrap properly?
While reviewing other tests of the visualisation backends I noticed that `TestSlice: test_equivalent_regions` in `test_visualisation.py` seems to be showing some symptoms of a possible bug, probably related to incorrect nearest neighbour finding when periodic wrapping of the box is involved. These are currently "silenced" with `frac={"nearest_neighbours": 0.75}.get(backend, 1.0),` as an argument to the `fraction_within_tolerance` helper function. This should be looked at to decide whether there really is a bug, and if there is, fix it.

kyleaoman:
There was an off-by-one error in the placing of pixel coordinates in the nearest-neighbour backend for slice visualisations. This is fixed, and the tests are adjusted to be more stringent (I had previously relaxed the tolerance on how well supposedly equivalent images should match for this backend as it "sort of" worked).

Closes #274

### DIFF
--- a/swiftsimio/visualisation/slice_backends/nearest_neighbours.py
+++ b/swiftsimio/visualisation/slice_backends/nearest_neighbours.py
@@ -1,6 +1,6 @@
 """Backend tools for image slices with nearest neighbour interpolation."""
 
-from numpy import float64, float32, ndarray, linspace, array, stack, meshgrid
+from numpy import float64, float32, ndarray, arange, array, stack, meshgrid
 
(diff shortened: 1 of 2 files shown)


robjmcgibbon (approved):
Nice catch, that's a subtle one!

kyleaoman:
I got lucky, one of the first things I looked at was the array of pixel coordinates, and it seemed suspicious that there were values only slightly bigger than 1.0...

PR #306, opened 2026-04-01
Expose rotation submodule

robjmcgibbon:
Rotations had to be imported explicitly, which I don't think should be required.

```
In [1]: import swiftsimio as sw
sw.
In [2]: sw.visualisation.rotation
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 sw.visualisation.rotation

AttributeError: module 'swiftsimio.visualisation' has no attribute 'rotation'

In [3]: import swiftsimio.visualisation.rotation
```

### DIFF
--- a/swiftsimio/visualisation/__init__.py
+++ b/swiftsimio/visualisation/__init__.py
@@ -4,6 +4,7 @@ from .projection import project_gas, project_pixel_grid
 from .slice import slice_gas
 from .volume_render import render_gas
 from .smoothing_length import generate_smoothing_lengths
+from . import rotation
 
 __all__ = [
     "project_gas",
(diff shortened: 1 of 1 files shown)


kyleaoman:
Fair enough. On the other hand, I've just had a look and this function is not used internally in swiftsimio itself anywhere. It is used:

- In the tests for rotated visualisations.
- In the documentation explaining how to make rotated visualisations.

The purpose is just to get the rotation matrix that aligns a vector with one of the coordinate axes. For that there's already `scipy.spatial.transform.Rotation.align_vectors`... I wonder if we could just drop this function, or make it a thin wrapper around that... anyway, for what it is this PR seems fine. Whenever you're ready just flag for review.

robjmcgibbon:
Thanks. I just had someone yesterday coming to ask me why it wasn't working, so I don't think we should get rid of it since some people must be relying on it.

kyleaoman (approved):

PR #310, opened 2026-04-08
Lazy-load field metadata

kyleaoman:
Read metadata for individual datasets at the same time as the data themselves.

This saves reading the units, cosmo_factor, compression, etc. for all datasets when initially opening the file, speeding up calls to `load()` (and `mask()`) substantially, especially for SOAP files with many datasets. Since most workflows do not involve reading all datasets in a file, this saves time overall.

### DIFF
--- a/swiftsimio/metadata/metadata/metadata_fields.py
+++ b/swiftsimio/metadata/metadata/metadata_fields.py
@@ -1,7 +1,7 @@
 """Define information needed to unpack metadata fields in SWIFT snapshots."""
 
 from unyt import Unit
-from ...objects import cosmo_factor
+from swiftsimio.objects import cosmo_factor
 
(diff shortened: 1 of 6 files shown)


swiftsimio/metadata/metadata/metadata_fields.py
```diff
@@ -1,7 +1,7 @@
 """Define information needed to unpack metadata fields in SWIFT snapshots."""
 
 from unyt import Unit
-from ..objects import cosmo_factor
+from ...objects import cosmo_factor
```
JBorrow: Triple dot is crazy, let's just do `from swiftismio.objects import cosmo_factor`

JBorrow (approved):
Looks good!

PR #311, opened 2026-04-08
Use the h5py low level API when reading dataset slices

### LINKED ISSUE #283: Subset writer tests are surprisingly slow
While testing #281 I found that test_subset_write.py is much faster when working on remote files accessed via the hdfstream web service. We seem to incur a lot of overhead in h5py's dataset.read_direct() method. Using local files on Cosma the test on EagleDistributed.hdf5 takes 70 seconds and spends 90% of that time in Dataset.read_direct(). Using remote files it takes 25 seconds, of which 70% is in SSLSocket.read(). 

Here's a profile of the case where we read local HDF5 files:

Image

read_direct() is called about 5000 times. I was able to reduce the runtime from 70 seconds to 20 seconds by using the h5py low level API to select all of the slices then read them with a single H5Dread(). But I'm not sure how this will affect more realistic cases and it results in a different ordering of the result if the slices are not sorted, so one of the SOAP tests fails.

The tests in test_mask.py show similar behaviour: running on remote files takes less than half as long. Although the individual tests are only ~5 seconds or less in that case.

jchelly:
Instead of doing a separate read_direct() for each slice when cutting out a region, we can select all slices with H5Sselect_hyperslab() and then do a single H5Dread(). This might avoid some h5py overhead. See also #283.


jchelly:
The test failure here seems to be due to overlapping ranges being selected in the test that tries to write a spatial subset of a SOAP catalogue. Will need to investigate why that happens and what the result should be.

jchelly:
Relevant information if we do try to use hyperslab selections: https://forum.hdfgroup.org/t/quadratic-runtime-of-selecting-n-hyperslabs/12555 .

PR #313, opened 2026-04-14
Add support for LOS files

### LINKED ISSUE #298: Unable to load los files
I have tried to open a line-of-sight (los) file with the swiftsimio.load function but swiftsimio does not recognize the filetype of the file.

Opening a random los file

```python
file = "/cosma8/data/dp004/colibre/Runs/L100_m6/Thermal/los/los_0231.hdf5"
data = swiftsimio.load(file)
```

gives the following error message:

`ValueError: File type LineOfSight not recognised.`

I have inspected the files attributes and everything is the same a simulation snapshot file. The problem is that within a los snapshot file, there are multiple line-of-sight fields created (LOS_xxxx) with xxxx going from 0000 up to 0102 for example. 

```python
file = "/cosma8/data/dp004/colibre/Runs/L100_m6/Thermal/los/los_0231.hdf5"
data = swiftsimio.load(file)
first_los = data.LOS_XXXX
```

Is there an easy way to let the reader function of swiftsimio open this filetype? Maybe with an extra flag so that it directly opens the LOS_XXXX field of choice?

robjmcgibbon:
This PR adds support for loading line of sight files. They are very similar to snapshot files, but contain a number of groups with the name `LOS_XXXX` (where `XXXX` ranges from `0000` to the number of lines of sight in the file). Each `LOS_XXXX` contains a number of gas fields (just like `PartType0` normally does), and holds all the gas particles that overlap with that line of sight.

To read the files I have added a new class called `SWIFTLineOfSightMetadata`. Each LOS file has 135 groups, so they benefit massively from the recently implemented lazy loading of metadata. Each group contains some attributes (the location of the line of sight), and so I am lazily loading those too. They can be accessed similarly to how the datasets are accessed (e.g. `data.los_0000.xpos`)

Tests are failing since they now require the file `ColibreLOS.hdf5`. This is just a copy of `/cosma8/data/dp004/colibre/Runs/L0012N0094/Thermal/los/los_0231.hdf5`, so could that please be copied to the web storage @kyleaoman?

@stankortmann is going to use this with the specwizard code, so he may have suggestions for some changes. Once people are happy with the implementation then I can add some documentation.


kyleaoman:
I've added the file, you'll have to update `webstorage_location` to `.../ssio_ci_04_2026` in `conftest.py`.

swiftsimio/reader.py
```diff
@@ -224,6 +224,77 @@ def setter(self: __SWIFTGroupDataset, value: cosmo_array) -> None:
...
+        if current_value is not None:
+            return current_value
+
+        with self.open_file() as handle:
+            value = handle[group].attrs[attr_name]
+
+        unit_loader = self.metadata.get_group_attribute_units(name)
+        unit = unit_loader(self.metadata.units)
+        comoving = self.metadata.get_group_attribute_comoving(name)
+        cf = self.metadata.get_group_attribute_cosmo_factor(
```
kyleaoman: Don't need the scale factor as an argument here, get it from `self` within the function.

swiftsimio/reader.py
```diff
@@ -224,6 +224,77 @@ def setter(self: __SWIFTGroupDataset, value: cosmo_array) -> None:
...
+            name, self.metadata.scale_factor
+        )
+
+        if np.ndim(value) == 0:
+            parsed = cosmo_quantity(value, unit, comoving=comoving, cosmo_factor=cf)
```
kyleaoman: This is a bit convoluted. I think you could get the same without all of the branching with:
```python
parsed = cosmo_array(np.atleast1d(value), ...).squeeze()
```

kyleaoman (commented):
Looks sane overall to me, couple of little suggestions.

swiftsimio/reader.py
robjmcgibbon: Thanks, done in https://github.com/SWIFTSIM/swiftsimio/pull/313/commits/901d2a2d3fd24617bca763459ace2b259b8246a9

swiftsimio/reader.py
robjmcgibbon: Also done in https://github.com/SWIFTSIM/swiftsimio/pull/313/commits/901d2a2d3fd24617bca763459ace2b259b8246a9

MatthieuSchaller:
Thanks, @robjmcgibbon, @stankortmann, and @kyleaoman. This will be very useful for many projects.

kyleaoman:
@robjmcgibbon this is still flagged draft, is it actually ready to merge, though? (Other than the conflicts...)

kyleaoman:
Conflicts seem to be around the subset_writer changes and should be easy to resolve. I suppose that a masked LoS file isn't a thing (note: could be spatial or by index) and therefore doesn't make sense to support in the subset writer? If that's wrong then should probably add a test that the subset writer actually works.

robjmcgibbon:
I don't think there's any use case for masking. There is already `masking_valid: bool = False` in `SWIFTLineOfSightMetadata`. Do you know if anything else is needed to be compatible with your latest changes? If not then I think this is ready

kyleaoman:
Just had a look, indeed setting `masking_valid = False` should be all that's needed.

kyleaoman (approved):

kyleaoman:
Hm, do you want to add a mention in the narrative docs?

robjmcgibbon:
Thanks, I've added a description. I don't think it's worth having a separate section for these files, so I've just put it at the bottom of https://swiftsimio.readthedocs.io/en/latest/loading_data/index.html

robjmcgibbon:
If you think it reads ok then I will merge

kyleaoman:
Looks good to me, thanks!

PR #315, opened 2026-04-23
Better comoving-physical conversion syntax

### LINKED ISSUE #314: Visualisation methods mutate the image region
When the image region is defined in physical coordinates, something is going wrong and the region (a `cosmo_array`) is being modified in-place, so trying to make the same image twice gives different results. This doesn't happen if the image region is in comoving coordinates. Here's a minimal reproduction:
```python
import matplotlib.pyplot as plt
import unyt as u
from swiftsimio import load, mask, cosmo_array
from swiftsimio.visualisation import project_gas

snapfile = "test_data/EagleSingle.hdf5"
m = mask(snapfile)
a = m.metadata.scale_factor
boxsize = m.metadata.boxsize
mask_region = cosmo_array([boxsize * 0, boxsize * 0.2]).T
m.constrain_spatial(mask_region)
d = load(snapfile, mask=m)
im_region = cosmo_array(
    [0.1, 0.2, 0.1, 0.2], u.Mpc, comoving=False, scale_factor=a, scale_exponent=1
)
im1 = project_gas(
    d,
    128,
    region=im_region,
    periodic=False,
)
im2 = project_gas(
    d,
    128,
    region=im_region,
    periodic=False,
)

plt.imsave("im1.png", im1)
plt.imsave("im2.png", im2)
```
The two images should be identical, but I get:

Image

Image

According to Angel, who brought this to my attention, consecutive images with "the same" region like this are progressively more "zoomed out".

A quick look at the visualisation tools reveals plenty of use of `convert_to_comoving`/`convert_to_physical`/`convert_to_units` which modify arrays in-place, so the likely cause is clear enough.

kyleaoman:
This refactors some `cosmo_array` functions for more flexible conversion between comoving and physical quantities. As a summary, these previously unavailable syntax options now work:
``` python
arr.to_comoving(u.kpc)
arr.to_physical(u.kpc)
arr.to_comoving_value(u.kpc)
arr.to_physical_value(u.kpc)
arr.to(u.kpc, comoving=False)
arr.to_value(u.kpc, comoving=False)
arr.convert_to_comoving(u.kpc)
arr.convert_to_physical(u.kpc)
```
This also leads to performance improvements where possible, by avoiding a second copy. For example this would have copied twice: `arr.to(u.kpc).to_physical()` while `arr.to_physical(u.kpc)` only copies once.

The docs are updated to reflect this, and I've added some more tests for these methods.

There's also some housekeeping around the `valid_transform` flag, which I think must have been added after the big extension of `cosmo_array` a while back.

Finally, it closes #314... this PR might have happened because I got sidetracked while being more careful about using in-place conversions in the visualisation tools. Possibly.

### DIFF
--- a/swiftsimio/_array_functions.py
+++ b/swiftsimio/_array_functions.py
@@ -192,8 +192,10 @@ def _propagate_cosmo_array_attributes_to_result(func: Callable) -> Callable:
     """
 
     def wrapped(
-        obj: object, *args: tuple[Any], **kwargs: dict[str, Any]
-    ) -> object:  # # noqa numpydoc ignore=GL08
+        obj: object,
+        *args: tuple[Any],
+        **kwargs: dict[str, Any],
+    ) -> object:  # noqa numpydoc ignore=GL08
         # omit docstring so that sphinx picks up docstring of wrapped function
         return _copy_cosmo_array_attributes_if_present(obj, func(obj, *args, **kwargs))
 
@@ -886,9 +888,9 @@ def _prepare_array_func_args(
     -------
     dict
         A dictionary containing the input ``args`` and ``kwargs`` coerced to a common
-        state, and lists of their ``cosmo_factor`` attributes, and ``comoving`` and
-        ``compression`` values that can be used in return values for wrapped functions,
-        when relevant.
+        state, and lists of their ``cosmo_factor`` attributes, and ``comoving``,
+        ``valid_transform`` and ``compression`` values that can be used in return values
+        for wrapped functions, when relevant.
 
     Raises
     ------
@@ -897,6 +899,7 @@ def _prepare_array_func_args(
     """
     cms = [(hasattr(arg, "comoving"), getattr(arg, "comoving", None)) for arg in args]
     cfs = [getattr(arg, "cosmo_factor", None) for arg in args]
+    vts = [getattr(arg, "valid_transform", True) for arg in args]
     comps = [
         (hasattr(arg, "compression"), getattr(arg, "compression", None)) for arg in args
     ]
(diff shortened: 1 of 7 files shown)


swiftsimio/_array_functions.py
```diff
@@ -955,20 +959,22 @@ def _prepare_array_func_args(
...
-    )
+    return {
+        "args": args,
+        "kwargs": kwargs,
+        "cfs": cfs,
+        "kw_cfs": kw_cfs,
+        "comoving": ret_cm,
+        "valid_transform": ret_vt,
+        "compression": ret_comp,
+    }
```
JBorrow, line 977: Why change this?

swiftsimio/_array_functions.py
```diff
@@ -1650,13 +1665,14 @@ def _prepare_array_block_args(lst: list, recursing: bool = False) -> dict:
...
-    return dict(
-        args=ret_lst,
-        kwargs=dict(),
-        comoving=ret_cm,
-        cosmo_factor=ret_cf,
-        compression=ret_comp,
-    )
+    return {
+        "args": ret_lst,
+        "kwargs": dict(),
```
JBorrow: For consistency "kwargs": {}

JBorrow (approved):
Very cool addition, no notes.

swiftsimio/_array_functions.py
kyleaoman: Seems to be a rule of thumb based on style and performance considerations, e.g. https://stackoverflow.com/questions/17097985/dict-vs-in-python-which-is-better, https://stackoverflow.com/questions/2853683/what-is-the-preferred-syntax-for-initializing-a-dict-curly-brace-literals-or, https://www.thepythoncodingstack.com/p/python-dict-is-more-versatile-than-you-may-think, etc, etc.

PR #316, opened 2026-04-30
Support for SIDM aka. PartType7

MatthieuSchaller:


MatthieuSchaller:
Better when done on top of a fresh master...

PR #318, opened 2026-05-01
Add an entry for new `sidm` particle type to required fields for writer

kyleaoman:
Pretty much as the subject says, this was missed when the new type was added. I added a test to catch similar issues in the future.

@JBorrow this breaks the whole swiftgalaxy test suite which relies on the writer, so perhaps a patch release? I don't think it's totally breaking things: I don't think it breaks the writer itself, or swiftgalaxy itself, I think I have a loop over particles types that isn't happy. But anyway, trivial fix.


JBorrow:
Sorry for missing that. I'll be happy to put out a patch release.

PR #319, opened 2026-05-01
Write consistent metadata for snapshot/soap subsets

### LINKED ISSUE #245: Particle counts not updated creating a snapshot subset
Whenever writing a subset of a snapshot as explained in [`subset_writer`](https://swiftsimio.readthedocs.io/en/latest/masking/index.html#writing-subset-of-snapshot), all of the original metadata aside from [cell counts](https://github.com/SWIFTSIM/swiftsimio/blob/9f78bb7ed325adc7968b34937f47ff639b377081/swiftsimio/subset_writer.py#L163) is simply copied over to the new snapshot.

This means that if someone loads the resulting subset snapshot, the _actual_ number of particles no longer matches the values given in `Header/NumPart_ThisFile` and `Header/NumPart_Total`. I believe those values should be updated when saving the subset snapshot to reflect the actual values. 

Before providing a fix, I wanted to confirm whether this behaviour is intented or not.

### LINKED ISSUE #312: `write_subset` fails without a spatial mask
The `subset_writer` refuses to write output if no spatial (cell) mask is provided in the mask object, but there are now reasonable use cases where this could happen. In particular, using `constrain_indices` and then wanting to write out the resulting subset seems reasonable. Assuming that no spatial mask just means "the whole box" is probably reasonable, although we should be careful to prevent "rewrite the whole snapshot/SOAP catalogue as a new file" which is probably not intended. Could add a check that a constraint of some kind has been applied to the mask, for example, which would just need adding a boolean flag to the mask object that gets flipped to `True` when any of the constrain functions are called.

As an illustration of the issue, here's a bit of sample code:
```python
soapfile = "/cosma8/data/dp004/colibre/Runs/L025_m7/Thermal/SOAP-HBT/halo_properties_0127.hdf5"
from swiftsimio import *
m.constrain_indices([1,2,3])
d = load(soapfile, mask=m)
subset_writer.write_subset("test.hdf5", m)  # fails: RuntimeError: Subset writing requires specifying a cell mask. Please use constrain_spatial with a suitable restrict array to generate one.
m = mask(soapfile)
region = np.vstack([m.metadata.boxsize * 0, m.metadata.boxsize]).T
m.constrain_spatial(region)
m.constrain_indices([1,2,3])
d = load(soapfile, mask=m)
subset_writer.write_subset("test.hdf5", m)  # this works fine!
```
The second (working) call to `write_subset` produces an output file that has only 3 rows in each dataset, which is what we expected, so writing the subset of the data seems to work fine. The output file can also be read in with `swiftsimio` and the datasets can be read successfully - so far so good.

On closer inspection, however, the metadata is not updated (see also #245 ). This includes "header" data like `metadata.num_part`, but also more crucially the cell metadata, which still thinks that all particles (or subhaloes, for SOAP) are present. Trying to use a `swiftsimio` mask on the output file would go poorly.

kyleaoman:
Now ready for review, this is a re-working of the `subset_writer` to write output with correct metadata (including the cell metadata). This also meant a substantial re-work of `masks`. I'm pinging Josh, Rob & John for reviews, I think any of you could do this, just a matter of who might want to pick it up.

- Implementing this generally requires more control over what masks can be doing: we need to be able to back out the spatial mask counts and offsets for the subset based on the mask for the full dataset at any time. I've therefore done a substantial internal refactor of the `SWIFTMask` class.
- `spatial_only` is no longer really that that label means, it's whether the mask is stored as ranges or as a boolean mask. That's been re-labeled `range_mask`.
- Enabled bi-directional conversion between bool and range masks (i.e. the old `constrain_spatial` `True` and `False`).
- These internal changes mean that users mostly don't need to think about the type of mask (conversion can be triggered automatically internally), but can still switch between boolean- and range-based masks with `convert_masks_to_bool` and `convert_masks_to_ranges` if needed. I've therefore deprecated the `spatial_only` kwarg to `mask()` (it still exists, just does nothing and produces a warning if used).
- (Sidenote: I also deprecated `intersect` argument to `constrain_spatial` in favour of `union` which actually describes what it does.)
- I've reworked how masks can be combined by repeated calls of e.g. `constrain_spatial`, `constrain_indices`, etc. I think that the best approach is to refuse to mask when `self.constrained` (a new attribute tracking whether a constraint has been applied) is not `None` except for specifically allowed exceptions (e.g. `constrain_spatial(..., union=True)`, also `constrain_mask`). `self.constrained` keeps a record of what the last applied mask function was as a string (or `None`, initially).
- `constrain_indices` has changed slightly, the data are always loaded in the order they were stored in the file (so a mask [1, 0] gets converted internally to [0, 1]). Input is checked for being sorted and users are warned of the changed ordering if they pass in unsorted input. Previously the ordering depended on whether `spatial_only` was `True` or `False`, so this seems to have less risk of confusion.
- Implemented working out the cell mask to write to a subset file for any arbitrary mask (boolean or range, doesn't matter).
- Currently `constrain_index`/`constrain_indices` is implemented so that the index is *always* the index into the full, unmasked snapshot/soap file, and cannot be applied after other masks. Are we happy with this, or is e.g. masking a region with `constrain_spatial` then narrowing down to some rows in that region with `constrain_indices` (this implies accepting a relative index, presumably, otherwise would have just used `constrain_indices` in the first place) considered a necessary feature?
- Added some new tests and reviewed the docstrings and docs.
- Since I was doing a pretty substantial refactor anyway I had a look for other more minor changes, e.g. some functions have been converted to `@property` where this made sense, found a few lines that were unused, etc. A few of these changes were guided by `mypy` (type checking) which I've been finding helpful in other projects, `masks.py` and `subset_writer.py` would now pass type checking, for whatever that's worth...
- Renamed `constrain_mask` to `constrain_property` (the old name is still there and functional with a deprecation warning).
- Restored the old ability to use `None` as a row in a mask region, this was no longer functioning (though still in the docs).

Closes #245
Closes #312


docs/source/loading_data/index.rst
```diff
@@ -356,7 +356,7 @@ place of the filename.
    # The full metadata object is available from within the mask
    boxsize = mask.metadata.boxsize
    # load_region is a 3x2 list [[left, right], [bottom, top], [front, back]]
-   load_region = [[0.0 * b, 0.5 * b] for b in boxsize]
+   load_region = np.vstack((0.0 * boxsize, 0.5 * boxsize)).T
```
JBorrow: Is this necessary? I think it's significantly less readable.

docs/source/masking/index.rst
```diff
@@ -220,31 +224,52 @@ the region can be extended or switched off with the ``safe_padding`` parameter:
...
-   mask.constrain_spatial([
-       [0.2 * mask.metadata.boxsize[0], 0.7 * mask.metadata.boxsize[0]],
-       None,
-       None]
+   mask.constrain_spatial(
+       np.array(
+           [
+	       [0.2, 0.7],
+	       [0.0, 1.0],
+	       [0.0, 1.0],
```
JBorrow: I don't really like the fact taht this needs to be an array now. It's way more annoying to construct.

A common use-case I have is selecting regions along the z-axis for 'slice' style images. So I always do something like `[None, None, [0.0 * unyt.Mpc, 50.0 * unyt.Mpc]]` which I'd like to be assumed to be co-moving. Which is nice to type and construct. I think right now I'd need to do something like...
```
cosmo_array(
  [
    [0.0, data.metadata.boxsize[0].to_comoving("Mpc")],
    [0.0, data.metadata.boxsize[1].to_comoving("Mpc")],
    [0.0, 50.0]
  ],
  units="Mpc",
  scale_exponent=1,
  scale_factor=data.metadata.boxsize.scale_factor,
```
I had to go look up the documentation just to construct this...

docs/source/masking/index.rst
```diff
@@ -310,22 +336,24 @@ as follows
...
+    mask = sw.mask("eagle_snapshot.hdf5")
+    boxsize = mask.metadata.boxsize
+    scale_factor = mask.metadata.scale_factor
+    mask.constrain_spatial(
+        np.vstack(
+	    (
+                cosmo_array([[100, 200]], u.kpc, comoving=True, scale_factor=scale_factor, scale_exponent=1),
+		np.array([[0.0, 0.1], [0.0, 0.1]]) * boxsize[1:, np.newaxis].to(u.kpc)
+	    )
+	)
```
JBorrow: I guess this is the same thing as above. It's just horrible developer ergonomics.

JBorrow (commented):
Haven't reviewed the actual implementation but the developer ergonomics of the new system are not good and we should re-think this.

kyleaoman:
Fair, but I'm actually just documenting the current behaviour:
```python
>>> from swiftsimio import mask
>>> m = mask("/cosma8/data/dp004/colibre/Runs/L0012N0094/Thermal/snapshots/colibre_0127/colibre_0127.hdf5")
>>> region = [np.array([0, 1]) * m.metadata.boxsize[0], None, None]
>>> region
[cosmo_array([ 0.        , 12.50000001], 'Mpc', comoving='True', cosmo_factor='a at a=0.9999999999999997', valid_transform='True'), None, None]
>>> m.constrain_spatial(region)
  File "<stdin>", line 1, in <module>
  File "/cosma/apps/durham/dc-oman1/code/swiftsimio/swiftsimio/masks.py", line 662, in constrain_spatial
    self.cell_mask = self._generate_cell_mask(restrict)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/cosma/apps/durham/dc-oman1/code/swiftsimio/swiftsimio/masks.py", line 518, in _generate_cell_mask
    lower = restrict[dimension][0]
            ~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```
Doesn't mean we shouldn't improve this.

kyleaoman:
@JBorrow should be all set now. One of the test runs is taking ages, but just seems to be a slow download speed on the test files, so nothing to worry about in the grand scheme.

docs/source/loading_data/index.rst
```diff
@@ -355,7 +355,7 @@ place of the filename.
    mask = sw.mask(snap_file)
    # The full metadata object is available from within the mask
    boxsize = mask.metadata.boxsize
-   # load_region is a 3x2 list [[left, right], [bottom, top], [front, back]]
+   # load_region is a 3x2 list [[left, right], [front, back], [bottom, top]]
```
JBorrow: `y` is 'front-back' and `z` is 'bottom-top'? I think the original comment is correct, no?

kyleaoman: `y` is front-back and `z` is bottom-top, as the *new* comment says? Confused...

docs/source/masking/index.rst
```diff
@@ -310,22 +344,22 @@ as follows
...
-        [unyt.unyt_quantity(100, unyt.kpc), unyt.unyt_quantity(1000, unyt.kpc)], 
-        None, 
-        None])                                   
-    
+    mask = sw.mask("eagle_snapshot.hdf5")
+    scale_factor = mask.metadata.scale_factor
+    mask.constrain_spatial(
+        [
+            cosmo_array([100, 200], u.kpc, comoving=True, scale_factor=scale_factor, scale_exponent=1),
+	    None,
```
JBorrow: Mixing of tabs and spaces?

kyleaoman: Yup - and because apparently I never do things by halves, I just ran `sphinx-lint` and fixed everything flagged (trailing whitespace, mixed tabs & spaces, line length, mostly), and added it to the CI pipeline.

swiftsimio/subset_writer.py
```diff
@@ -9,19 +9,19 @@
 import numpy as np
 
 
-def get_swift_name(name: str) -> str:
+def _get_swift_name(name: str) -> str:
```
JBorrow, line 12: This is generally fine but I don't think we need to be making every little thing a private function as it unecessarily hides things.

JBorrow: I guess my reasoning here is also that we're introducing breaking API changes for no good reason. I know someone probably wasn't using get_swift_name outside of this functionality, but if they were, they're now broken. Same as changing the argument name from 'recurse' to 'recursing' below.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ packages = [
 
 [project]
 name = "swiftsimio"
-version="12.1.1"
+version="12.1.2"
```
JBorrow, line 25: This includes breaking API changes so should be a major revision.

kyleaoman: Current master is 12.1.3, I don't know why this commit is here (it's actually your commit) as I rebased on master... Anyway, yes, this should be a major version bump on release.

swiftsimio/subset_writer.py
kyleaoman: I guess my counterargument is that if as a user I open up the docs for `subset_writer` I'm confronted with a bunch of not-obviously-related stuff cluttering around what I actually want, which is just `write_subset`. Arguably the whole thing could be refactored into a class as there's a lot of passing information back and forth that could make sense as class attributes, but I decided not to go that far.

kyleaoman: If you prefer reverting this it's not hard to do, though.

On the `recurse`/`recursing`, I initially tried to refactor this out because I don't see why initializing the empty list explicitly on the first call is needed (can't you just call it with an empty list for `dataset_names` on the first call?). Trying to remove it immediately broke things so I put it back (with a new name, it seems), but I don't think I understand what's going on - is it clear to you why that flag is even needed?

JBorrow: I think it's because you've gotta set the default value of the dataset_names' default correctly, otherwise you modify the default list:
```
>>> def my_func(x: list[str] = []):
...   x.append("a")
...   return x
...
>>> my_func()
['a']
>>> my_func()
['a', 'a']
>>> my_func()
['a', 'a', 'a']
>>> my_func()
['a', 'a', 'a', 'a']
>>> my_func()
['a', 'a', 'a', 'a', 'a']
```
If we're not changing functionality in that function specifically, you should revert the change I think. There's no need for the churn.

JBorrow (approved):
Happy aside from reverting the `recursing` thing.

PR #320, opened 2026-05-04
Conversion of arrays with no valid transform

robjmcgibbon:
Changes in https://github.com/SWIFTSIM/swiftsimio/pull/315 mean it is no longer possible to call `.to` on objects without a valid transform.
```
import swiftsimio as sw
import unyt as u

cq = sw.cosmo_quantity(
    1,
    u.m,
    comoving=False,
    scale_factor=1.0,
    scale_exponent=0,
    valid_transform=False,
)

cq.to('Mpc')
```
gives
```
  File "/cosma/home/dp004/dc-mcgi1/swiftsimio/example.py", line 13, in <module>
    cq.to('Mpc')
  File "/cosma/home/dp004/dc-mcgi1/swiftsimio/swiftsimio/objects.py", line 1654, in to
    raise InvalidConversionError
swiftsimio.objects.InvalidConversionError
```
This is due to a check that should be skipped if `comoving=None` (i.e. we don't want to convert the array)

### DIFF
--- a/docs/source/citing/index.rst
+++ b/docs/source/citing/index.rst
@@ -1,5 +1,5 @@
-Citing SWIFTGalaxy
-==================
+Citing swiftsimio
+=================
 
 .. include:: ../../../README.rst
    :start-after: CITING_START_LABEL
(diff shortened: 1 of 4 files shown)


kyleaoman (approved):
I guess this needs a release?

robjmcgibbon:
Thanks, yes I think it would be good to put out a patch release

PR #328, opened 2026-05-15
Add helpers to more easily create `cosmo_array` and `cosmo_quantity`

### LINKED ISSUE #297: Improve UI for creating `cosmo_array`
This is a proposal for a new feature to make creating `cosmo_array` (and related `cosmo_quantity`) objects much easier. 

### Context

The code has provided the data as `cosmo_array` objects since its earliest versions. These have 3 metadata fields relevant to this discussion: the units, a comoving/physical flag and how the quantity scales (like lengths go as `a**1`). Relatively recently the `cosmo_array` has been fleshed out into a full-fledged subclass of `ndarray` and `unyt_array`, which in practice means that you can feed these arrays through any `numpy` and many `numpy`-like functions and any necessary unit and comoving <-> physical conversions happen seamlessly, there are extra consistency checks, etc.

The problem that I think needs addressing is that to make full use of the `cosmo_array` means that you should create new `cosmo_array` or `cosmo_quantity` (for scalars) objects. For reference if you have a `unyt_array`, let's say it's called `distances`, and you want to compare to a scalar, then the user-friendly syntax is `distances < 5 * u.Mpc`. Currently to do this correctly with a `cosmo_array` of distances, you would write something like:
```python
distances < cosmo_quantity(
    5,
    u.Mpc,
    comoving=True,
    scale_factor=distances.scale_factor,
    scale_exponent=1, 
)
```
This seems like a lot to ask and many users probably compare to `unyt_array` or `unyt_quantity` (using the more convenient syntax illustrated above) instead. The question then is what kind of sensible syntax could be offered to make using the "cosmo" objects easier. Much of the point of having this machinery in place is not to assume defaults for the units, `comoving` flag, `scale_factor`  and `scale_exponent`. This needs to be balanced against the need for a concise, easy to use syntax. 

### Proposed syntax

After a bit of discussion on the SWIFT Slack, we have a proposal for a new syntax to create `cosmo_array` and `cosmo_quantity` objects. 

```python
dat = load(...)  # load a swift dataset as usual
a = dat.metadata.a
10 * u.Mpc * a.comoving**1
10 * u.Mpc * a.physical**1
```

We currently have both `dat.metadata.a` and `dat.metadata.scale_factor` that are aliases for the same thing (a `float`). Under this proposal `dat.metadata.a` would become a new helper object instead. Inspired by the `unyt` multiplication syntax this can be used to "cosmo-ify" objects.

We cannot assume that lengths always scale as `a**1` or generally infer the scaling exponent from the dimensions of the units, so the exponent syntax is used for this. In this example `**1` is written explicitly but this would not be required, an exponent of `1` would be the implicit default so `10 * u.Mpc * a.comoving` is equivalent. Also, we would provide examples and encourage assigning to variables to create more concise shorthands, for instance this would be valid:
```python
cMpc = u.Mpc * dat.metadata.a.comoving
10 * cMpc
```
or, also valid:
```python
acom = dat.metadata.a.comoving
10 * u.Mpc * acom
```

### Request for input

For the moment set aside technical considerations around how this would be implemented. Would you be likely to use this syntax? Do you have any alternatives to propose? Any concerns with the proposed syntax?

kyleaoman:
As discussed in #297 this PR adds a new helper to make initialising `cosmo_array` and `cosmo_quantity` objects easier. It's assigned to the name `dat.metadata.a`, with the bare scale factor still available as `dat.metadata.scale_factor`. The basic usage is what we converged on in discussion, e.g.
```python
m = mask("dat.hdf5")
acom = m.metadata.a.comoving  # or whatever name you want
region = [[10 * acom * u.Mpc, 20 * acom * u.Mpc], None, None]
m.constrain_spatial(region)
```
To avoid bugs in migrating to the new system, trying to use `dat.metadata.a` as if it was still a `float` will raise an error:
```python
10 * dat.metadata.a * u.Mpc  # raises
```
This is achieved by ensuring that either `dat.metadata.a.comoving` or `dat.metadata.a.physical` has been used. The documentation will encourage people to assign to a short alias.

The one exception is that through a little clever trick the `cosmo_array` and `cosmo_quantity` (and `cosmo_factor`) constructors are backwards compatible, this still works:
```python
cosmo_quantity(1, u.Mpc, comoving=True, scale_factor=dat.metadata.a, scale_exponent=1)
```
That will hopefully help not completely break all existing scripts on upgrading `swiftsimio`.

Still to do:
- [x] Fill out error messages.
- [x] Fill out/check docstrings.
- [x] Write & update narrative docs.
- [x] Add some more tests, I haven't covered division yet.
- [x]  `*=`, `/=`, `**=` should be tested. Don't want to implement with helper as left argument, that seems dangerous/confusing (shouldn't mutate the helper attached to the metadata!), as right argument it should already work through `__rmul__`. 
- [x] Check if there is any orphaned logic in the `_AHelper` class, coverage analysis should help with this. I'm not totally sure that all of the methods are actually used & therefore needed, `numpy` is a bit cagey with redirecting things into `__array_ufunc__`.
- [ ] Also attach helper to `data.a`, in addition to `data.metadata.a`.
- [ ] This requires changes on the `unyt` that completely block us when a `unyt.Unit` or `unyt.unyt_array` is the left-hand argument, unfortunately. Tracking https://github.com/yt-project/unyt/pull/640 & https://github.com/yt-project/unyt/pull/674. This branch currently set to test with my branch of unyt, that needs reverting if/when those PRs are merged.

### DIFF
--- a/docs/source/cosmo_array/index.rst
+++ b/docs/source/cosmo_array/index.rst
@@ -101,6 +101,9 @@ cases a warning is produced stating that this assumption has been made.
    supported (although some might "just work"). Requests to support specific functions can
    be accommodated.
 
+Creating "cosmo" arrays and scalars
+-----------------------------------
+
 To make the most of the utility offered by the :class:`~swiftsimio.objects.cosmo_array`
 class, it is helpful to know how to create your own. A good template for this looks like:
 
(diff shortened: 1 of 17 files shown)


kyleaoman:
@JBorrow this could be looked at now (in which case make sure to grab my fork of unyt if you want to run tests locally, see `pyproject.toml`). Or, could wait on the upstream changes.

This one turned into a bit of a monster as getting the helper working properly needed some housekeeping around typing, decided I may as well do this properly but tried to keep it mostly limited to `objects.py` and `_array_functions.py`.

Suggest starting with the `_AHelper` class itself, the docstrings there give lots of context.

kyleaoman:
Managed to find a much simple approach on the unyt side with lots of resulting simplification on the swiftsimio side. Still need to wait for upstream changes, but a discussion has started now.

JBorrow (approved):

PR #329, opened 2026-05-18
Easy cloning of `cosmo_array` attributes with `like` kwarg

### LINKED ISSUE #297: Improve UI for creating `cosmo_array`

kyleaoman:
`numpy` offers the opportunity for array subclasses to intervene when creating new array through the `like` keyword argument. This PR implements this feature for `cosmo_array` (and its subclass `cosmo_quantity`) allowing us to create new arrays that copy the `units`, `comoving`, `cosmo_factor`, `compression` and `valid_transform` attributes from an existing array. In other words, this very nice (imo) syntax now works:

```python
>>> from swiftsimio import load
>>> dat = load("test_data/ColibreSingle.hdf5")
>>> distances = np.arange(3, like=dat.gas.coordinates)
>>> distances
cosmo_array([0, 1, 2], 'Mpc', comoving='True', cosmo_factor='a**1.0 at a=0.5', valid_transform='True')
```

The following functions are supported:
 - `np.asarray`
 - `np.asanyarray`
 - `np.ascontiguousarray`
 - `np.asfortranarray`
 - `np.require`
 - `np.fromfunction`
 - `np.fromstring`
 - `np.fromiter`
 - `np.fromfile`
 - `np.frombuffer`
 - `np.identity`
 - `np.loadtxt`
 - `np.genfromtxt`
 - `np.eye`
 - `np.tri`

Still to-do:
 - [x] Updated narrative documentation.
 - [x] Check some of the main docstrings for examples that could be udpated.

Helps address #297 (see also #328 for the rest of the changes discussed in that issue). Had originally implemented this in the other PR but it's a standalone change and doesn't depend on changes in `unyt` so makes sense to split it off into its own PR.

### DIFF
--- a/docs/source/cosmo_array/index.rst
+++ b/docs/source/cosmo_array/index.rst
@@ -21,8 +21,8 @@ The cosmology information is stored in three attributes:
  + ``cosmo_factor``
  + ``valid_transform``
 
-The ``comoving`` attribute specifies whether the array is a physical (``True``) or
-comoving (``False``) quantity, while the ``cosmo_factor`` stores the expression needed
+The ``comoving`` attribute specifies whether the array is a physical (``False``) or
+comoving (``True``) quantity, while the ``cosmo_factor`` stores the expression needed
 to convert back and forth between comoving and physical quantities and the value of
 the scale factor. The conversion factors can be accessed like this:
 
@@ -122,6 +122,24 @@ class, it is helpful to know how to create your own. A good template for this lo
    # with:
    # scale_factor=data.metadata.a
 
+This can be tedious to write out, but often a new array with the same units, ``comoving``,
+and scale factor & exponent as an existing :class:`~swiftsimio.objects.cosmo_array` is
+wanted. All of the main :mod:`numpy` array creation functions such as :func:`numpy.array`,
+:func:`numpy.asarray` and several others accept a keyword argument called ``like``. Using
+this with a :class:`~swiftsimio.objects.cosmo_array` will copy over all of the metadata
+but use the new values provided, for example:
+
+.. code-block:: python
+
+   my_new_cosmo_array = np.asarray(
+      [4, 5, 6],
+      like=my_cosmo_array,
+   )
+   # has units of u.Mpc, comoving=True, scale_factor=0.5, scale_exponent=1
+
+Keep in mind that :func:`numpy.asarray` tries to avoid copying the input data when
+possible while :func:`numpy.array` usually makes a copy with a corresponding memory cost.
+
 There is also a very similar :class:`~swiftsimio.objects.cosmo_quantity` class designed
 for scalar values, analogous to the :class:`~unyt.array.unyt_quantity`. You may encounter
 this being returned by :mod:`numpy` functions. Cosmology-aware scalar values can be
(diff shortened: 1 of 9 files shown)


kyleaoman:
@JBorrow this is really neat and intuitive, I think, and a pretty straightforward addition.

JBorrow (approved):
Sorry that I took so long to take a look at this. I had imagined it would be a horribly complex implementation, but it seems fairly straightforward! Thanks for this change, it is really neat.

kyleaoman:
Yes indeed this is the easy case.

PR #330, opened 2026-05-27
Use human-friendly time units in swiftsnap output

smsutherland:
Just a small quality-of-life change with swiftsnap. Since the time unit comes from the length and velocity units, the actual value of the unit can be a little inconvenient to work with. All this PR does is change the time unit to be more human-friendly when printing in swiftsnap.
For example:
`Simulation state: z=0, a=1, t=0.01437 977.79*Gyr`
becomes
`Simulation state: z=0, a=1, t=14.05 Gyr`


kyleaoman (approved):

kyleaoman:
Nice and straightforward. Thanks for the contribution!

PR #333, opened 2026-06-01
Support non-gas slices and renders

robjmcgibbon:
Ian McCarthy asked if it was possible to make slices of dark matter. We already support projections for non-gas particles via ``project_pixel_grid``, but slices and volume renders were gas-only. This PR extends both to support any particle type by introducing ``slice_pixel_grid`` and ``render_voxel_grid``. The existing ``slice_gas`` and ``render_gas functions`` are now wrappers.

The tests now call ``slice_pixel_grid`` and ``render_voxel_grid`` using dark matter. That might be overkill though since it's not testing anything different to the gas case, so I think I'll remove that.

Also fixes a silent bug in the nearest_neighbours smoothing length backend, which was always falling back to SPH smoothing lengths due to double ``.gas`` attribute access.

### DIFF
--- a/swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
+++ b/swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
@@ -23,13 +23,12 @@ def get_hsml(data: __SWIFTGroupDataset) -> cosmo_array:
     cosmo_array
         The extracted "smoothing lengths".
     """
-    try:
+    if hasattr(data, "volumes"):
         hsml = np.cbrt(data.volumes)
-    except AttributeError:
-        try:
-            # Try computing the volumes explicitly?
-            hsml = np.cbrt(data.masses / data.densities)
-        except AttributeError:
-            # Fall back to SPH behavior if above didn't work...
-            hsml = get_hsml_sph(data)
+    elif hasattr(data, "masses") and hasattr(data, "densities"):
+        # Try computing the volumes explicitly?
+        hsml = np.cbrt(data.masses / data.densities)
+    else:
+        # Fall back to SPH behavior if above didn't work...
+        hsml = get_hsml_sph(data)
     return hsml
--- a/swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
+++ b/swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
@@ -1,11 +1,12 @@
 """Tools to get nearest-neighbour smoothing lengths."""
 
 import numpy as np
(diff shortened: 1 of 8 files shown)


swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
```diff
@@ -14,20 +15,20 @@ def get_hsml(data: SWIFTDataset) -> cosmo_array:
...
 
     Returns
     -------
     cosmo_array
         The extracted "smoothing lengths".
     """
     try:
-        hsml = np.cbrt(data.gas.volumes)
+        hsml = np.cbrt(data.volumes)
     except AttributeError:
```
kyleaoman: Suggest rewriting these two `try` blocks with `hasattr` or `getattr`. `except AttributeError` is a bit indiscriminate, makes it hard to debug or even notice when say something internal in the field getter throws an `AttributeError`.

swiftsimio/visualisation/volume_render.py
```diff
@@ -85,11 +86,11 @@ def render_gas(
 
     See Also
     --------
-    slice_gas_pixel_grid
-        Creates a 2D slice of a SWIFT dataset.
+    slice_pixel_grid
```
kyleaoman, line 91: Also project_pixel_grid?

kyleaoman (changes requested):
Looks mostly sensible to me, and definitely makes sense to offer equivalent support for all of the visualisation tools.

swiftsimio/visualisation/smoothing_length/nearest_neighbours.py
robjmcgibbon: Good point thanks, done in https://github.com/SWIFTSIM/swiftsimio/pull/333/commits/51749bc47837b79cbf785e4fac205647b527e9c5

swiftsimio/visualisation/volume_render.py
robjmcgibbon: Done in https://github.com/SWIFTSIM/swiftsimio/pull/333/commits/51749bc47837b79cbf785e4fac205647b527e9c5

robjmcgibbon:
Tests are failing as the test data can't currently be downloaded, presumably related to the cosma downtime

kyleaoman (approved):

kyleaoman:
Tests pass locally (except unrelated ones needing `ColibreLOS.hdf5`, which I don't have). Seems ready to merge when cosma comes back up, or sooner I guess if Ian was looking to use this "right now".

PR #334, opened 2026-06-05
Ensure consistent return type in numba function

kyleaoman:
Fixes an issue where numba inferred an int-only return type in one conditional branch but a float-or-int return type in the other branch. This is an error. Since this function is returning indices for arrays I've changed it to explicitly return an int type array.


jchelly (approved):
Thanks Kyle, that looks good to me.

jchelly:
The "blame" display on github doesn't seem to be working in this branch. Looks like git (or github?) wants a full length hash - see #335.

PR #335, opened 2026-06-08
Use full commit hash in .git-blame-ignore-revs

jchelly:
While looking at #334 I found that the "blame" option on github is not working. It complains that the .git-blame-ignore-revs file is invalid. Using a full length commit hash seems to fix it.


kyleaoman (approved):
Had noticed this issue but never looked into it further, thanks for finding a simple fix!

PR #336, opened 2026-06-09
Masking on slice and volume render

smsutherland:
This PR adds the mask kwarg to `slice_pixel_grid` and `slice_gas`, behaving similarly to the equivalent kwarg in `project_pixel_grid` and `project_gas`. I've also added a test for the functionality. I didn't see any tests for the mask kwarg for projections, so I include a test for that as well.

### DIFF
--- a/tests/test_visualisation.py
+++ b/tests/test_visualisation.py
@@ -1604,3 +1604,81 @@ def test_input_region_unmutated(cosmological_volume_only_single_local):
     assert np.allclose(im_region.to_physical_value(unyt.Mpc), float_region)
     project_gas(data, 16, region=im_region, periodic=False)
     assert np.allclose(im_region.to_physical_value(unyt.Mpc), float_region)
+
+
+class TestVisualisationMask:
+    """Tests for the particle masking functionality in projections and slices."""
+
+    def test_projection(self, cosmological_volume_only_single_local):
+        """
+        Tests masking for projections.
+
+        Passing the mask parameter should be equivalent to filtering out the un-masked data and calling without a mask.
+        """
+        sd = load(cosmological_volume_only_single_local)
+        n_gas = sd.metadata.n_gas
+
+        mask = np.zeros(n_gas, dtype=np.bool)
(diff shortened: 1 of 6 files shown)


smsutherland:
I suppose it wouldn't be that hard to extend this PR to volume renders as well, to bring all three into parity in this regard, if that's something y'all would be interested in.

kyleaoman:
Thanks for the contribution! Looks good to me so far. I agree that it would make sense to extend to `volume_render` too, if you don't mind adding that? Should be straightforward like the others.

Did you look at the narrative docs at all? Might be worth a brief mention of the feature there.

smsutherland:
There's volume rendering done. I'll see if I can get something useful written for the narrative docs.

tests/test_visualisation.py
```diff
@@ -1604,3 +1604,124 @@ def test_input_region_unmutated(cosmological_volume_only_single_local):
...
+                masked_arr = getattr(sd.gas, f)[::2]
+                setattr(sd.gas, f, masked_arr)
+
+            ref_img = render_gas(
+                sd,
+                resolution=box_res,
+                parallel=parallel,
+                periodic=True,
+            )
+            assert np.allclose(masked_img, ref_img)
```
kyleaoman, line 1729: Small suggestion for all of the tests, also worth asserting that there's at least one non-zero element in the images being compared to avoid getting a trivial "false pass".

smsutherland: Good catch!

smsutherland (commented):
Commit [d6549cf](https://github.com/SWIFTSIM/swiftsimio/pull/336/commits/d6549cfffb3dad91e49fd593bbcf86dbe109720a) isn't strictly relevant to the rest of the PR. I figured I'd fix the typo while I was there.
(dang it I thought I commented on the commit itself.)

kyleaoman (approved):
Looks good to me, @smsutherland are you ready for this to merge?

smsutherland:
Ready if you are

PR #341, opened 2026-09-05
Fix statistics field identifiers

### LINKED ISSUE #89: load_statsitics can have ( in variable names

Ryukon07:
The statistics reader currently turns `SFR (total)` into `sfr_(total)`, which can only be reached through `getattr`. This change normalizes statistics headings into valid Python identifiers while preserving the existing dot, space, and lowercase behavior.

It also handles leading digits, Python keywords, and headings containing only punctuation. The new end-to-end test reads a temporary statistics file and verifies that `SFR (total)` is available as `statistics.sfr_total`; the statistics documentation now describes the mapping.

Validation:
- `pytest -q tests/test_statistics.py`
- `ruff check .`
- `ruff format --check .`
- `python -m numpydoc lint swiftsimio/statistics.py tests/test_statistics.py`
- `sphinx-lint --max-line-length 90 -e all docs/source/`
- `git diff --check`

Fixes #89


swiftsimio/statistics.py
```diff
@@ -1,11 +1,43 @@
...
+
+    Returns
+    -------
+    str
+        Normalized name suitable for attribute access.
+    """
+    # Keep the established normalization for dots and spaces so existing
+    # attribute names remain stable, then handle all other invalid characters.
+    clean_name = name.replace(".", "").replace(" ", "_").strip().lower()
+    clean_name = re.sub(r"[^a-zA-Z0-9_]", "_", clean_name)
```
JBorrow, line 28: Wouldn't this middle regex replace the above replacement anyway?

